PR #84, opened 2025-06-06
Story/FCT2 6606 : Copy Confirmation

renjithabby:


ui-spa/src/components/case-management/transfer-materials/NetAppFolderContainer.tsx
```diff
@@ -249,6 +295,43 @@ const NetAppFolderContainer: React.FC<NetAppFolderContainerProps> = ({
     handleGetFolderContent(path);
   };
 
+  const getInsetElement = () => {
+    const curentFolder = folders[folders.length - 1];
+    return (
+      <InsetText data-testId="netapp-inset-text">
```
HasanCPS, line 301: same here

PR #86, opened 2025-06-18
Feature/fct 6521 transfer status

bridgesr:


backend/CPS.ComplexCases.API/Functions/Transfer/GetTransferStatus.cs
```diff
@@ -0,0 +1,37 @@
...
+    [OpenApiResponseWithBody(statusCode: HttpStatusCode.OK, contentType: ContentType.ApplicationJson, bodyType: typeof(string), Description = ApiResponseDescriptions.Success)]
+    [OpenApiResponseWithBody(statusCode: HttpStatusCode.BadRequest, contentType: ContentType.TextPlain, typeof(string), Description = ApiResponseDescriptions.BadRequest)]
+    [OpenApiResponseWithBody(statusCode: HttpStatusCode.Unauthorized, contentType: ContentType.TextPlain, typeof(string), Description = ApiResponseDescriptions.Unauthorized)]
+    [OpenApiResponseWithBody(statusCode: HttpStatusCode.Forbidden, contentType: ContentType.TextPlain, typeof(string), Description = ApiResponseDescriptions.Forbidden)]
+    [OpenApiResponseWithBody(statusCode: HttpStatusCode.InternalServerError, contentType: ContentType.TextPlain, typeof(string), Description = ApiResponseDescriptions.InternalServerError)]
+    public async Task<IActionResult> Run(
+        [HttpTrigger(AuthorizationLevel.Anonymous, "get", Route = "filetransfer/{transferId}/status")] HttpRequest req, FunctionContext functionContext, string transferId)
```
HasanCPS, line 29: The function is currently using `AuthorizationLevel.Anonymous`. For security, this should be changed to `AuthorizationLevel.Function` to ensure only requests with the correct function key can access this endpoint.

backend/CPS.ComplexCases.API/Functions/Transfer/GetTransferStatus.cs
```diff
@@ -0,0 +1,37 @@
...
+{
+    private readonly ILogger<GetTransferStatus> _logger = logger;
+    private readonly IFileTransferClient _transferClient = transferClient;
+
+    [Function(nameof(GetTransferStatus))]
+    [OpenApiOperation(operationId: nameof(Run), tags: ["File Transfer"], Description = "Get status of a file transfer.")]
+    [OpenApiParameter(name: "caseId", In = ParameterLocation.Path, Required = true, Type = typeof(string), Description = "The Id of the transfer.")]
+    [OpenApiResponseWithBody(statusCode: HttpStatusCode.OK, contentType: ContentType.ApplicationJson, bodyType: typeof(string), Description = ApiResponseDescriptions.Success)]
+    [OpenApiResponseWithBody(statusCode: HttpStatusCode.BadRequest, contentType: ContentType.TextPlain, typeof(string), Description = ApiResponseDescriptions.BadRequest)]
+    [OpenApiResponseWithBody(statusCode: HttpStatusCode.Unauthorized, contentType: ContentType.TextPlain, typeof(string), Description = ApiResponseDescriptions.Unauthorized)]
```
HasanCPS, line 25: There is no explicit handling for authentication errors. If the user is not authenticated or provides invalid credentials, the function should return a `401 Unauthorized` status code.
Please ensure that authentication failures (e.g., `CpsAuthenticationException`) are caught and return a `401 Unauthorized `response.

backend/CPS.ComplexCases.API/Functions/Transfer/GetTransferStatus.cs
bridgesr: The authentication check is done in the RequestValidationMiddleware. The gateway api is called from the UI so it makes no sense to pass the azure function api token here (would be exposed on the client) the Function level auth is in the File Transfer API

backend/CPS.ComplexCases.API/Functions/Transfer/GetTransferStatus.cs
```diff
@@ -29,10 +30,21 @@ public class GetTransferStatus(ILogger<GetTransferStatus> logger, IFileTransferC
...
+        {
+            var context = functionContext.GetRequestContext();
 
-        var response = await _transferClient.GetFileTransferStatusAsync(transferId, context.CorrelationId);
+            var response = await _transferClient.GetFileTransferStatusAsync(transferId, context.CorrelationId);
 
-        return await response.ToActionResult();
+            return await response.ToActionResult();
+        }
+        catch (CpsAuthenticationException)
```
HasanCPS, line 41: I tested this locally and found that when I don't provide an auth token in the request, I'm getting a 500 Internal Server Error instead of the expected 401 Unauthorized.

Root Cause: The issue is that our middleware (`RequestValidationMiddleware.cs`) is throwing the `CpsAuthenticationException`, but the function's `try-catch` block never gets executed because the middleware runs before the function in the pipeline:

Request → Middleware (throws exception) → Function (never reached)


Since the function never executes, the catch block that's supposed to return a 401 response is never hit, and the Azure Functions runtime converts the unhandled exception to a 500 error.

Suggested Fix: We should handle the authentication error directly in the middleware instead of throwing an exception. This would ensure consistent 401 responses for auth failures across all endpoints.

HasanCPS (approved):
LGTM!

PR #94, opened 2025-07-01
fct2_7681 enforce transfer options

bridgesr:


backend/CPS.ComplexCases.API/Validators/Requests/InitiateTransferRequestValidator.cs
```diff
@@ -13,5 +14,19 @@ public InitiateTransferRequestValidator()
         RuleFor(x => x.SourcePaths).Must(x => x.Count > 0).WithMessage("At least one SourcePath is required.");
```
HasanCPS, line 14: Currently only validates that the collection is not empty, but doesn't validate individual `SourcePath `objects. This means empty or invalid paths could pass validation.

Suggestion: Add validation for individual `SourcePath `items:

```
RuleFor(x => x.SourcePaths).Must(x => x.Count > 0).WithMessage("At least one SourcePath is required.");
RuleForEach(x => x.SourcePaths).ChildRules(sourcePath => {
    sourcePath.RuleFor(x => x.Path).NotEmpty().WithMessage("SourcePath.Path is required.");
});
```

HasanCPS (approved):
LGTM!

PR #104, opened 2025-07-11
Feature/fct2 7679 implement move functionality in filetransfer

NeilFoubister:


backend/CPS.ComplexCases.Data/Repositories/ActivityLogRepository.cs
```diff
@@ -70,6 +70,10 @@ public async Task<IEnumerable<ActivityLog>> GetByFilterAsync(ActivityLogFilterDt
             query = query.Where(a => a.ResourceId == filter.ResourceId);
         }
 
-        return await query.Skip(filter.Skip).Take(filter.Take).OrderByDescending(x => new { x.Timestamp, x.Id }).ToListAsync();
+        return await query
+            .Skip(filter.Skip)
+            .Take(filter.Take)
+            //.OrderByDescending(x => new { x.Timestamp, x.Id })
```
HasanCPS, line 76: I noticed that the `OrderByDescending(x => new { x.Timestamp, x.Id })` line has been commented out. Removing explicit ordering can lead to inconsistent results when paginating, since the database may return records in a non-deterministic order. This could cause users to see duplicate or missing records across pages.

**Suggestion:**  
If there isn’t a specific reason for removing the ordering, I recommend keeping the `OrderByDescending` clause to ensure consistent and predictable pagination. If performance is a concern, we could look into optimizing the query or adding appropriate indexes.

Let me know your thoughts or if there’s a particular reason for this change—happy to discuss further!

backend/CPS.ComplexCases.FileTransfer.API/Durable/Activity/DeleteFiles.cs
```diff
@@ -0,0 +1,71 @@
...
+
+    [Function(nameof(DeleteFiles))]
+    public async Task Run([ActivityTrigger] DeleteFilesPayload payload, CancellationToken cancellationToken = default)
+    {
+        if (payload == null)
+        {
+            throw new ArgumentNullException(nameof(payload), "DeleteFilesPayload cannot be null.");
+        }
+
+        if (payload.TransferDirection == Common.Models.Domain.Enums.TransferDirection.NetAppToEgress)
```
HasanCPS, line 26: This check hard-codes a single invalid direction, but the orchestration logic (see `TransferOrchestrator.cs`) only calls this activity for `EgressToNetApp `and `Move` transfers. If new directions are added in the future, this could silently allow unsupported cases. Consider using a whitelist of allowed directions or validating against the orchestrator's logic for maintainability. 

For example,

```
// At the top of the class, define allowed directions (could also be static readonly)
private static readonly HashSet<TransferDirection> AllowedDirections = new()
{
    TransferDirection.EgressToNetApp
    // Add more allowed directions here as needed
};

...

if (!AllowedDirections.Contains(payload.TransferDirection))
{
    _logger.LogError("Unsupported transfer direction for DeleteFiles activity: {TransferDirection}", payload.TransferDirection);
    throw new ArgumentException($"Unsupported transfer direction for DeleteFiles activity: {payload.TransferDirection}", nameof(payload));
}
```

backend/CPS.ComplexCases.Data/Repositories/ActivityLogRepository.cs
HasanCPS: Thanks for the explanation! It would be helpful to add a comment in the code explaining why the ordering was commented out and that you plan to revisit it as part of the `ActivityLog `updates. This will provide context for anyone else looking at the code in the meantime.

backend/CPS.ComplexCases.FileTransfer.API.Tests/Unit/Durable/Activity/DeleteFilesTests.cs
```diff
@@ -0,0 +1,203 @@
...
+        await Assert.ThrowsAsync<InvalidOperationException>(() =>
+            _activity.Run(payload, CancellationToken.None));
+    }
+
+    [Fact]
+    public async Task Run_LogsAndReturns_WhenNoFilesToDelete()
+    {
+        // Arrange
+        var transferId = _fixture.Create<Guid>();
+        var expectedErrorMessage = $"No files to delete for transfer ID {transferId}.";
```
HasanCPS, line 96: The test checks for a log message containing this string, but the actual log message in the implementation is parameterized. If the log message format changes, this test will break. Consider using a more flexible assertion or extracting the log message as a constant in the implementation for reuse.

backend/CPS.ComplexCases.FileTransfer.API.Tests/Unit/Functions/ListFilesForTransferTests.cs
```diff
@@ -212,6 +217,32 @@ public async Task Run_NetAppToEgress_SkipsPathValidation()
         Assert.True(filesResult.ValidationErrors == null || !filesResult.ValidationErrors.Any());
     }
 
+    [Fact]
+    public async Task Run_MoveFrom_EgressToNetApp_WithUserPermissionCheck_ReturnsEgressPermissionExceptionResult_WhenUserLacksPermission()
```
HasanCPS, line 221: The test covers the negative permission path. Consider adding a positive test to ensure that when permission is granted, the function does not return an exception result and proceeds as expected.

PR #115, opened 2025-07-21
Devops/cleanup agent after deployment

lilachdavis:
Add step to clean up downloaded build artifacts after deployment


devops-pipelines/backend/backend-deploy.yml
```diff
@@ -254,6 +256,17 @@ stages:
                       fi 
 
                       echo "✅ Database health check completed successfully"
+
+                - script: |
```
HasanCPS, line 260: Would it be possible to create a `cleanup.sh` script (or similar) and reuse it? It looks like this cleanup logic is being used in multiple places.

HasanCPS (approved):
LGTM!

PR #132, opened 2025-07-29
fix: logging information level

bridgesr:


backend/CPS.ComplexCases.API/Program.cs
```diff
@@ -26,11 +26,32 @@
...
+                configureApplicationInsightsLoggerOptions: (options) => { }
+            );
+        }
+
+        if (context.HostingEnvironment.IsDevelopment())
+        {
+            logging.AddConsole();
+        }
+
+        logging.SetMinimumLevel(LogLevel.Information);
```
HasanCPS, line 53: Great job on the logging setup! One suggestion: consider making the minimum log level configurable via `local.settings.json` or Azure App Settings, rather than hardcoding it. This will give us more flexibility to adjust log verbosity across different environments without code changes, and aligns with best practices for cloud-native applications. Thanks!

backend/CPS.ComplexCases.API/Program.cs
```diff
@@ -26,11 +26,32 @@
 
 var host = new HostBuilder()
     .ConfigureFunctionsWebApplication(webApp =>
-    { 
+    {
         // note: the order of middleware is important, as it determines the execution flow
         webApp.UseMiddleware<ExceptionHandlingMiddleware>();
         webApp.UseMiddleware<RequestValidationMiddleware>();
     }) // ✅ Adds ASP.NET Core integration
+    .ConfigureLogging((context, logging) =>
```
HasanCPS, line 34: Could we also add that for the `CPS.ComplexCases.FileTransfer.API`?

HasanCPS (approved):
LGTM!

PR #154, opened 2025-08-27
Story/FCT2-9052- Private Beta Group Access Error

renjithabby:
https://cpsgovuk.atlassian.net/browse/FCT2-9052


ui-spa/src/auth/msal/AccessErrorPage.tsx
```diff
@@ -0,0 +1,26 @@
...
+    <Layout>
+      <div className="govuk-width-container">
+        <div className={styles.contentWrapper}>
+          <h1 className="govuk-heading-xl">Access Error</h1>
+          <p>
+            You cannot access this page. You are not a member of this group.
+          </p>
+          <p>
+            If you think this is a mistake, contact{" "}
+            <a href="mailto:meadhbh.major@cps.gov.uk">
```
HasanCPS, line 15: Thanks for putting this together — it’s looking really solid 👏.
Just one small suggestion: instead of hardcoding a specific email address directly here, can we reference it from a config variable (e.g., environment variable)? That way, if the contact ever changes, we won’t need a code change to update it.

This also helps us avoid exposing PII directly in the repo. What do you think about pulling it from config?

HasanCPS (approved):
LGTM!

PR #159, opened 2025-09-09
Devops/deployment slots

HasanCPS:


devops-pipelines/templates/backend-deploy-stages.yml
```diff
@@ -175,42 +178,28 @@ stages:
...
+        steps:
+          - template: ./fa-deploy-steps.yml
+            parameters:
+              environment: ${{ parameters.environment }}
+              azureSubscription: ${{ parameters.azureSubscription }}
+              functionAppName: $(functionAppName)
+              pipelineResourceAlias: 'build'
+              buildArtifactName: $(buildArtifactName)
+              deployToSlot: true
+              slotName: 'stg'
```
HasanCPS, line 201: Noticed that the deployment slot name is currently hardcoded in the pipeline as `stg`. I'd suggest making it configurable so we can pass in the appropriate slot name per environment, based on our agreed naming conventions. Since we might not use the same slot name across all environments, this flexibility would be useful. I understand there are still some pending discussions around this, so in the meantime, it would be great to make it easy to update.

devops-pipelines/templates/ui-build-and-deploy-stages.yml
```diff
@@ -109,10 +114,59 @@ stages:
                   resourceGroupName: $(resourceGroupName)
                   appType: 'webAppLinux'
                   WebAppName: '$(webAppName)'
+                  deployToSlotOrASE: true
```
HasanCPS, line 117: We can reuse the same approach we used for FA above by referencing `${{ parameters.deployToSlot }}`. This will make the slot configurable and consistent across environments.

lilachdavis (approved):

lilachdavis (approved):

PR #326, opened 2026-03-18
FCT2-14888: Add Playwright E2E tests for LCC file transfer workflows

kmshrajcps:
Egress to NetApp Copy  - Copy a single file from Egress to NetApp 
 Egress to NetApp Copy - Large  Copy a large file from Egress to NetApp 
 Egress to NetApp Copy - Copy multiple files from Egress to NetApp 
 Egress to NetApp Move - Move files from Egress to NetApp (skipped) 
 NetApp to Egress Copy - Copy a file from NetApp to Egress
 Login Only -Full login flow and workspace/case setup verification

Capture6
Capture7
Capture8
Capture9


e2e/pw/pipelines/e2e-tests.yml
HasanCPS, line 1: Do we need this file? I would suggest that Lilach adds it separately once the PR is merged, since it’s within her area of expertise.

e2e/pw/playwright.config.ts
```diff
@@ -0,0 +1,31 @@
...
+dotenv.config({
+  path: path.resolve(__dirname, `.env.${process.env.ENVIRONMENT || "local"}`),
+});
+
+export default defineConfig({
+  testDir: "./tests",
+  fullyParallel: true,
+  forbidOnly: !!process.env.CI,
+  retries: process.env.CI ? 2 : 0,
+  workers: 2,
```
HasanCPS: Each `setupTestData` run does full Egress workspace + upload + **`registerCase`** + login. Two workers ⇒ **two concurrent** full setups, doubling load on Egress/Case API/CMS and **flake** risk. If intentional for speed, OK; otherwise consider `workers: 1` in CI or document the trade-off.

e2e/pw/pages/ActivityLogTab.ts
```diff
@@ -0,0 +1,59 @@
...
+    await this.page
+      .getByTestId("activities-timeline")
+      .waitFor({ state: "visible", timeout: 30000 });
+  }
+
+  async getLatestActivity() {
+    return this.page
+      .getByTestId("activities-timeline")
+      .locator("section")
+      .first();
```
HasanCPS, line 25: `verifyTransferLogged` **filters** by description, but `expandFileList` / `downloadCsv` use **`section.first()`**, which may be a **connection** event if it appears first. Reuse the same filtered section as `verifyTransferLogged` or pass the target section in to avoid **wrong-row** clicks.

e2e/pw/pages/TransferMaterialsTab.ts
```diff
@@ -0,0 +1,118 @@
...
+    for (const index of indices) {
+      await checkboxes[index].check();
+    }
+  }
+
+  async selectNetAppFiles(indices: number[]) {
+    const checkboxes = await this.page
+      .locator("table tbody tr")
+      .locator('input[type="checkbox"]')
+      .all();
```
HasanCPS: `selectEgressFiles` uses `.locator("table").first()`; `selectNetAppFiles` uses unscoped `table tbody tr`. **ui-spa** Playwright uses `getByTestId("egress-table-wrapper")` / `netapp-table-wrapper`. Aligning reduces breakage if DOM order changes.

e2e/pw/pages/TransferMaterialsTab.ts
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  /** Wait until at least `expectedCount` file rows appear in the Egress panel folder, refreshing periodically. */
+  async waitForFileCount(expectedCount: number, folderName: string, timeout: number = 120_000) {
+    const egressTable = this.page.locator("table").first();
+    const start = Date.now();
+    // Check current count first (already navigated into folder)
+    let rows = await egressTable.locator("tbody tr").all();
+    if (rows.length >= expectedCount) return;
+
```
HasanCPS, line 255: Polling with **full `page.reload()`** every 5s is heavy and can hide client issues. Acceptable if Egress indexing is the bottleneck — add a one-line comment **why** reload is required so future readers do not remove it blindly.

e2e/pw/helpers/egress-api.ts
```diff
@@ -0,0 +1,232 @@
...
+      }
+    }
+
+    if (workspaces.length < pageSize) break;
+    page++;
+  }
+
+  // Add random suffix to avoid race conditions when running tests in parallel
+  const randomSuffix = Math.floor(Math.random() * 900 + 100);
+  return `AUTOMATION-TESTING${maxNumber + 1}-${randomSuffix}`;
```
HasanCPS, line 120: Scan regex is `^AUTOMATION-TESTING(\d+)$` but generated names are `` `AUTOMATION-TESTING${maxNumber + 1}-${randomSuffix}` ``. Hyphenated workspaces from this helper **do not** match the regex, so **`maxNumber` often stays 0** and names look like `AUTOMATION-TESTING1-{random}`. **Uniqueness** still works via the suffix; if monotonic `…TESTING2`, `3` was intended, fix the regex or naming. Please confirm current behaviour is OK.

e2e/pw/tests/egress-to-netapp-move.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+import { EgressConnectPage } from "../pages/EgressConnectPage";
+import { EgressConfirmationPage } from "../pages/EgressConfirmationPage";
+import { NetAppConnectPage } from "../pages/NetAppConnectPage";
+import { NetAppConfirmationPage } from "../pages/NetAppConfirmationPage";
+import { CaseManagementPage } from "../pages/CaseManagementPage";
+import { TransferMaterialsTab } from "../pages/TransferMaterialsTab";
+import { ActivityLogTab } from "../pages/ActivityLogTab";
+
+test.describe("Egress to NetApp Move", () => {
+  test.skip("should move files from Egress to NetApp", async ({
```
HasanCPS: The spec is **skipped**, but since we currently allow move functionality from lower environments (e.g., Dev and Staging), this should be fine for testing. We should enable it.

e2e/pw/tests/login-only.spec.ts
```diff
@@ -0,0 +1,252 @@
...
+
+  console.log("[7/7] Registering case...");
+  const { caseId, caseUrn } = await registerCase(
+    config.caseApiBaseUrl,
+    accessToken,
+    cmsAuth,
+    workspaceName
+  );
+  console.log(`  Case: ${caseUrn} (ID: ${caseId})`);
+  console.log("=== API Setup Complete ===\n");
```
HasanCPS: Reimplements the same API path as **`setup-helper.ts`** instead of the shared fixture — **two places** to fix when setup changes. Prefer extending the fixture or importing `setupTestData`.

e2e/pw/tests/login-only.spec.ts
```diff
@@ -0,0 +1,252 @@
...
+  await netappConnect.waitForFolders();
+  console.log("  NetApp folders displayed!");
+
+  // Click on the "Automation-Testing" folder's Connect button
+  console.log("  Clicking Automation-Testing folder...");
+  await page
+    .getByRole("row", { name: "Automation-Testing Connect", exact: false })
+    .filter({ hasNotText: "Move" })
+    .locator('button[name="secondary"]')
+    .click();
```
HasanCPS: Other specs use **`NetAppConnectPage.connectFolder(0)`**. This test matches **`Automation-Testing`** in the row text (Shared Drive folder label). That is **environment-specific**; naming/order changes → flake. Align with POM + index or document required NetApp folder names for CI.

e2e/pw/fixtures/setup-helper.ts
```diff
@@ -0,0 +1,149 @@
...
+  await page.goto(BASE_URL!);
+  const adLogin = new AzureADLoginPage(page);
+  await adLogin.login(E2E_AD_USER!, E2E_AD_PASSWORD!);
+  console.log("  Azure AD login complete.");
+
+  console.log("  Waiting for radio buttons...");
+  const caseSearch = new CaseSearchPage(page);
+  await caseSearch.waitForRadioButtons();
+  console.log("  All 3 radio buttons visible!");
+
```
HasanCPS, line 194: Every test using `testData` creates a **new** workspace + **new** `registerCase` + login. Full runs create **many** CMS cases and Egress workspaces. Optional env to **reuse** a pre-provisioned case (non-CI) would cut cost; document **parallel** hazards if you add it.

e2e/pw/pages/ActivityLogTab.ts
```diff
@@ -0,0 +1,59 @@
...
+    const latest = await this.getLatestActivity();
+    await latest
+      .getByRole("button", { name: /Download the list of files/i })
+      .click();
+  }
+
+  async verifyDownloadSuccess() {
+    // Download triggers a file download - just verify the button was clickable
+    // (no tooltip testid exists on the page)
+  }
```
HasanCPS, line 106: Empty body; comment says no tooltip testid. **`ui-spa`** has `activity-download-tooltip` in `ActivityTimeline.tsx` — if production exposes it, assert here; else tighten the comment **why** E2E skips it.

e2e/pw/pages/TransferMaterialsTab.ts
```diff
@@ -0,0 +1,118 @@
...
+    }
+  }
+
+  async selectAllNetAppFiles() {
+    await this.page.getByLabel("Select folders and files").check();
+  }
+
+  async openTransferDropdown(index: number = 0) {
+    // No-op: the Copy/Move buttons are directly visible in the inset text area
+  }
```
HasanCPS: Method is intentionally empty but specs still call it — confusing. Remove calls or remove the method.

e2e/pw/tests/egress-to-netapp-copy-multifile.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+    // Step 2: Click into case from results
+    const searchResults = new SearchResultsPage(page);
+    await searchResults.waitForResults();
+    await searchResults.clickCaseAction(caseUrn);
+
+    // Step 3: Connect Egress workspace
+    const egressConnect = new EgressConnectPage(page);
+    await egressConnect.searchFolder(workspace.name);
+    await egressConnect.waitForResults();
+    await egressConnect.connectFolder();
```
HasanCPS: `SearchResults.tsx` `getConnectOrViewUrl`: fully connected cases link to **`/case/{id}/case-management`** (“View”). Current E2E always runs egress + netapp connect after search, which matches **new** cases only. If you add **existing-case** reuse, branch on link or deep-link to case management.

e2e/pw/pipelines/e2e-tests.yml
kmshrajcps: file removed

e2e/pw/playwright.config.ts
kmshrajcps: updated to 1 worker

e2e/pw/pages/ActivityLogTab.ts
kmshrajcps: addressed

e2e/pw/pages/TransferMaterialsTab.ts
kmshrajcps: updated

e2e/pw/pages/TransferMaterialsTab.ts
kmshrajcps: comment added

e2e/pw/helpers/egress-api.ts
kmshrajcps: regex updated

e2e/pw/tests/egress-to-netapp-move.spec.ts
kmshrajcps: done

e2e/pw/tests/login-only.spec.ts
kmshrajcps: using shared test fixture

e2e/pw/tests/login-only.spec.ts
kmshrajcps: done

e2e/pw/fixtures/setup-helper.ts
kmshrajcps: *-default.spec.ts  uses  pre-existing case

e2e/pw/pages/ActivityLogTab.ts
kmshrajcps: assert added

e2e/pw/pages/TransferMaterialsTab.ts
kmshrajcps: removed

e2e/pw/tests/egress-to-netapp-move-default.spec.ts
```diff
@@ -0,0 +1,58 @@
+import { test, expect } from "../fixtures/test-fixtures-default";
+import { CaseSearchPage } from "../pages/CaseSearchPage";
+import { SearchResultsPage } from "../pages/SearchResultsPage";
+import { CaseManagementPage } from "../pages/CaseManagementPage";
+import { TransferMaterialsTab } from "../pages/TransferMaterialsTab";
+import { ActivityLogTab } from "../pages/ActivityLogTab";
+
+test.describe("Egress to NetApp Move (Default Mode)", () => {
+  test.skip("should move files from Egress to NetApp using existing case", async ({
```
HasanCPS: Why are we skipping this? It shouldn’t be skipped. Even though it’s currently disabled in Prod, we can still test it in our dev environment to ensure the functionality works as expected.

e2e/pw/.gitignore
HasanCPS, line 1: Instead of creating a new `.gitignore` file, can we use this existing one instead?
[https://github.com/CPS-Innovation/Large-and-Complex-Cases/blob/FCT2-14888/.gitignore](https://github.com/CPS-Innovation/Large-and-Complex-Cases/blob/FCT2-14888/.gitignore)

e2e/pw/tests/egress-to-netapp-copy-default.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+    const transferTab = new TransferMaterialsTab(page);
+    await transferTab.waitForEgressFiles();
+    await transferTab.navigateToFolder("4. Served Evidence");
+    await transferTab.waitForEgressFiles();
+
+    // Select the just-uploaded file by name to avoid picking old files already on NetApp
+    for (const file of testData.files) {
+      await transferTab.selectEgressFileByName(file.fileName);
+    }
+
```
HasanCPS: This spec-to-spec dependency via `test-results/.last-transferred-file.txt` is
brittle. It fails if the tests run out of order, in isolation, in parallel
shards, or after `test-results/` is cleaned. The silent fallback to
`testData.files[0].fileName` will succeed against a stale file on NetApp and
hide regressions.

Options:
  1. Make `netapp-to-egress-copy-default` self-sufficient by picking *any*
     recent file from the NetApp panel (e.g. sort by modified-date desc and
     take row 0), which is what the original `netapp-to-egress-copy.spec.ts`
     does.
  2. If coupling is intended, move both assertions into a single spec with
     two `test.step` blocks so ordering and state are explicit.

Please remove the shared file hand-off before merge.

e2e/pw/pages/TransferMaterialsTab.ts
```diff
@@ -0,0 +1,145 @@
...
+
+  async selectAllEgressFiles() {
+    await this.page.getByLabel("Select folders and files").check();
+  }
+
+  async selectEgressFiles(indices: number[]) {
+    const checkboxes = await this.page
+      .getByTestId("egress-table-wrapper")
+      .locator("tbody tr")
```
HasanCPS: These two methods have identical bodies and will match the same `aria-label`
on the Transfer Materials page. Playwright strict mode will throw when both
panels are visible, and without strict mode the NetApp variant will silently
click the Egress checkbox.

Please scope each to its panel, e.g.:

    async selectAllEgressFiles() {
      await this.page.getByTestId("egress-table-wrapper")
        .getByLabel("Select folders and files").check();
    }

    async selectAllNetAppFiles() {
      await this.page.getByTestId("netapp-table-wrapper")
        .getByLabel("Select folders and files").check();
    }

e2e/pw/pages/CaseSearchPage.ts
```diff
@@ -0,0 +1,43 @@
...
+    await this.page.getByTestId("radio-search-urn").check();
+  }
+
+  async fillUrn(urn: string) {
+    await this.page.getByTestId("search-urn").fill(urn);
+  }
+
+  async clickSearch() {
+    await this.page.locator("button").click();
+  }
```
HasanCPS, line 39: `page.locator("button")` is too broad and will either throw in strict mode or
click whichever button renders first (navigation, cookie banner, etc.). Use
an explicit selector, e.g. `getByRole("button", { name: "Search" })` or add a
`data-testid="search-submit"` to the underlying component.

e2e/pw/.env.template
```diff
@@ -0,0 +1,26 @@
...
+DDEI_ACCESS_KEY=                   # DDEI function key
+
+# Egress
+EGRESS_SERVICE_ACCOUNT_AUTH=       # Base64 service account credentials
+EGRESS_TEMPLATE_ID=59a6855307087630eb190282
+EGRESS_ADMIN_ROLE_ID=591dab08368b665c9c5c5fe0
+
+# Test Parameters
+TEST_FILE_SIZE_MB=100              # Test file size
+TEST_FILE_COUNT=1                  # Number of files to upload
```
HasanCPS, line 27: `.env.template` is missing the four `DEFAULT_*` variables that default mode
requires. A fresh engineer running `cp .env.template .env.local` cannot get
`npm run e2e:existing-case` to work without cross-referencing the README.

Please append to `.env.template`:

    # Default mode — required for npm run e2e:existing-case
    DEFAULT_WORKSPACE_ID=                # Pre-existing Egress workspace ID
    DEFAULT_WORKSPACE_NAME=              # Pre-existing Egress workspace name
    DEFAULT_CASE_URN=                    # Pre-existing case URN
    DEFAULT_CASE_ID=                     # Pre-existing case numeric ID

…and add a note above them explaining these are only needed for default
mode.

e2e/pw/helpers/env-config.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    egressServiceAccountAuth: requireEnv("EGRESS_SERVICE_ACCOUNT_AUTH"),
+    egressTemplateId: optionalEnv(
+      "EGRESS_TEMPLATE_ID",
+      "59a6855307087630eb190282"
+    ),
+    egressAdminRoleId: optionalEnv(
+      "EGRESS_ADMIN_ROLE_ID",
+      "591dab08368b665c9c5c5fe0"
+    ),
```
HasanCPS: These defaults hard-code tenant-specific Egress IDs. If they're non-sensitive
shared constants, please move them to a `constants.ts` with a comment
explaining why they are safe to commit.

e2e/pw/tests/egress-to-netapp-move-default.spec.ts
kmshrajcps: in dev we have only copy, move is not enabled 
Capture10

e2e/pw/.gitignore
kmshrajcps: removed gitignore and moved to the root gitignore

e2e/pw/tests/egress-to-netapp-move-default.spec.ts
HasanCPS: when did you check that

kmshrajcps: 10 mints ago

e2e/pw/helpers/egress-api.ts
```diff
@@ -0,0 +1,232 @@
...
+      `${baseUrl}/api/v1/workspaces/?page=${page}`,
+      { headers: { Authorization: `Basic ${token}` } }
+    );
+
+    if (!response.ok) break;
+
+    const data = await response.json();
+    const workspaces: { name: string }[] = data.data || data.results || data;
+
+    if (!Array.isArray(workspaces) || workspaces.length === 0) break;
```
HasanCPS: Please add a hard page cap (e.g. `if (page > 50) break;`) and drop the
`|| data`/`|| results` guessing in favour of the actual shape your API
documents. Also: pass the page size on the URL (`?page=${page}&page_size=50`)
so the client-side `pageSize` constant matches the server behaviour.

e2e/pw/helpers/egress-api.ts
```diff
@@ -0,0 +1,232 @@
...
+    );
+  }
+
+  // The file ID will be discovered later via the LCC API during tests.
+  // The Egress file listing API doesn't reliably expose subfolder contents,
+  // so we use the upload ID as reference and trust the upload completed.
+  console.log(`  Upload complete: ${uploadId}`);
+  return { fileId: uploadId, fileName, fileSize: fileSizeBytes };
+}
+
```
HasanCPS: `fileId` is actually the upload ID, not a file ID — the comment spells this
out but the type doesn't. Either:
  - rename the field to `uploadId` to match reality, or
  - drop it entirely from the return type (no consumer uses it today).

e2e/pw/fixtures/setup-helper.ts
```diff
@@ -0,0 +1,149 @@
...
+import { AzureADLoginPage } from "../pages/AzureADLoginPage";
+import { CaseSearchPage } from "../pages/CaseSearchPage";
+import type { TestSetupResult, UploadedFile } from "../helpers/types";
+
+export interface SetupOptions {
+  fileSizeMb?: number;
+  fileCount?: number;
+}
+
+export async function setupTestData(
```
HasanCPS, line 22: Please either:
  - add teardown fixtures that delete the workspace and (if possible) archive
    the registered case after the test, **or**
  - document the expected cleanup strategy (nightly script? support ticket?)
    in the README so ops aren't surprised by drift.

This is the single biggest operational risk in this PR — without cleanup, the
auto-numbered `AUTOMATION-TESTING{N}` workspaces will grow indefinitely.

e2e/pw/playwright.config.ts
```diff
@@ -0,0 +1,31 @@
...
+
+dotenv.config({
+  path: path.resolve(__dirname, `.env.${process.env.ENVIRONMENT || "local"}`),
+});
+
+export default defineConfig({
+  testDir: "./tests",
+  fullyParallel: true,
+  forbidOnly: !!process.env.CI,
+  retries: process.env.CI ? 2 : 0,
```
HasanCPS: `fullyParallel: true` and `workers: 1` contradict each other. Given the
README explains why single-worker is required, please set
`fullyParallel: false` (or delete the line) so the config reflects the
intended behaviour.

e2e/pw/pages/TransferMaterialsTab.ts
```diff
@@ -0,0 +1,145 @@
...
+    const successBanner = this.page.getByTestId("transfer-success-notification-banner");
+    const errorHeading = this.page.locator('text="There is a problem transferring files"');
+
+    // Wait for either success banner or error page to appear
+    await Promise.race([
+      successBanner.waitFor({ state: "visible", timeout }),
+      errorHeading.waitFor({ state: "visible", timeout }),
+    ]);
+
+    // Assert it was the success banner, not the error page
```
HasanCPS: Consider asserting the success banner visibly at the end of this method
(rather than relying on `errorHeading.isVisible()` being false). It's a
clearer intent signal and won't drift if the error heading becomes
asynchronous relative to the banner.

e2e/pw/tests/egress-to-netapp-copy-default.spec.ts
```diff
@@ -0,0 +1,67 @@
+import { test, expect } from "../fixtures/test-fixtures-default";
```
HasanCPS: `egress-to-netapp-copy.spec.ts`, `egress-to-netapp-copy-default.spec.ts`, `egress-to-netapp-copy-large.spec.ts`, `egress-to-netapp-copy-large-default.spec.ts`, `egress-to-netapp-copy-multifile.spec.ts`, `egress-to-netapp-move.spec.ts`, `egress-to-netapp-move-default.spec.ts`, `netapp-to-egress-copy.spec.ts`

They import `{ test, expect }` but never call `expect`. Either add an explicit `await expect(activityLog.getLatestActivity()).toBeVisible()` at the end (preferred, tightens the assertion surface) or drop the unused import.

These specs import `expect` but never call it — the whole test is implicit
assertions via `waitFor`. Consider:
  - removing the unused import to silence TS/ESLint, or
  - better, adding explicit `expect(...)` checks for the final activity-log
    entry so a failure mode like "logs panel rendered but transfer entry
    missing" is caught.

e2e/pw/pages/TransferMaterialsTab.ts
```diff
@@ -0,0 +1,145 @@
...
+      .locator('input[type="checkbox"]')
+      .all();
+    for (const index of indices) {
+      await checkboxes[index].check();
+    }
+  }
+
+  async selectNetAppFiles(indices: number[]) {
+    const checkboxes = await this.page
+      .getByTestId("netapp-table-wrapper")
```
HasanCPS: Prefer `nth(i)` on a fresh locator so each iteration re-queries the DOM:

    const rows = this.page.getByTestId("egress-table-wrapper").locator("tbody tr");
    for (const index of indices) {
      await rows.nth(index).locator('input[type="checkbox"]').check();
    }

Same pattern applies to `selectNetAppFiles`.

e2e/pw/helpers/env-config.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    testFileSizeMb: parseInt(optionalEnv("TEST_FILE_SIZE_MB", "100"), 10),
+    testFileCount: parseInt(optionalEnv("TEST_FILE_COUNT", "1"), 10),
```
HasanCPS: Add a `Number.isFinite`/`> 0` guard so a typo in `TEST_FILE_SIZE_MB` fails
fast with a clear error message rather than deep in the upload helper.

e2e/pw/helpers/types.ts
```diff
@@ -0,0 +1,22 @@
...
+export interface UploadedFile {
+  fileId: string;
+  fileName: string;
+  fileSize: number;
+}
+
+export interface TestSetupResult {
+  workspace: EgressWorkspace;
+  files: UploadedFile[];
+  caseId: number;
```
HasanCPS: `e2e/pw/helpers/types.ts:15`, `e2e/pw/helpers/env-config.ts:46`, `e2e/pw/fixtures/setup-helper-default.ts:42,114`

```ts
// types.ts
caseId: number;

// env-config.ts
defaultCaseId: process.env.DEFAULT_CASE_ID || "",   // string

// setup-helper-default.ts
const caseId = config.defaultCaseId;   // inferred as string
...
return { workspace: {...}, files, caseId, caseUrn };   // type error
```

`setupDefaultTestData` assigns a `string` to `caseId` but the return type declares `number`. Playwright's runtime transpiler (esbuild-based) does not type-check, so this does **not** throw at runtime today — but any pre-commit `tsc --noEmit` or CI type-check will fail. It also means `caseId` logged by `setup-helper-default.ts:85` may be the literal empty string `""` if `DEFAULT_CASE_ID` is unset.

`TestSetupResult.caseId: number` but `setupDefaultTestData` returns
`config.defaultCaseId` (string). Either:
  - change `caseId` to `number | string` in `TestSetupResult`,
  - `parseInt(process.env.DEFAULT_CASE_ID ?? "0", 10)` in `env-config.ts`,
  - or drop the field from `TestSetupResult` (nothing downstream uses it).

Please add `tsc --noEmit` as a lint step in `package.json` so this class of
bug gets caught pre-merge.

e2e/pw/fixtures/setup-helper-default.ts
```diff
@@ -0,0 +1,117 @@
...
+    console.log(`  Uploading ${fileName} (${i}/${fileCount})...`);
+    const file = await uploadFile(
+      config.egressBaseUrl,
+      egressToken,
+      workspaceId,
+      fileSizeBytes,
+      fileName
+    );
+    files.push(file);
+  }
```
HasanCPS, line 119: Even though default mode doesn't create new workspaces or cases, it *does* upload a fresh test file to the shared `DEFAULT_WORKSPACE_ID` on every invocation. After a week of CI + manual runs the `4. Served Evidence` folder fills up with hundreds of `generated-100MB-*-file1.txt` files. This also slows the "wait for file to be indexed" step in the large-file default spec because Egress has more files to enumerate.

Each default-mode run uploads a new file into the shared workspace with no
teardown. A long-running environment will accumulate thousands of test
files. Options:
  - add a teardown that deletes the just-uploaded file after the test,
  - upload to a dated subfolder and prune subfolders older than N days,

e2e/pw/tests/egress-to-netapp-move-default.spec.ts
HasanCPS: Are you sure? I can see the move.
Screenshot 2026-04-22 123143

e2e/pw/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "e2e-pw",
+  "version": "1.0.0",
+  "scripts": {
+    "e2e": "npx playwright test --ui",
+    "e2e:ci": "npx playwright test",
+    "e2e:existing-case": "npx playwright test --grep \"Default Mode\"",
```
HasanCPS: Can we add UI mode for the existing case?

For example: 

`"e2e:existing-case:ui": "npx playwright test --ui --grep \"Default Mode\"",`

e2e/pw/package.json
```diff
@@ -0,0 +1,21 @@
...
+    "e2e:register-case": "npx playwright test --grep-invert \"Default Mode\"",
```
HasanCPS: Same for the register case.

`"e2e:register-case:ui": "npx playwright test --ui --grep-invert \"Default Mode\"",`

e2e/pw/tests/egress-to-netapp-copy-large.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+import { SearchResultsPage } from "../pages/SearchResultsPage";
+import { EgressConnectPage } from "../pages/EgressConnectPage";
+import { EgressConfirmationPage } from "../pages/EgressConfirmationPage";
+import { NetAppConnectPage } from "../pages/NetAppConnectPage";
+import { NetAppConfirmationPage } from "../pages/NetAppConfirmationPage";
+import { CaseManagementPage } from "../pages/CaseManagementPage";
+import { TransferMaterialsTab } from "../pages/TransferMaterialsTab";
+import { ActivityLogTab } from "../pages/ActivityLogTab";
+
+test.describe("Egress to NetApp Copy - Large File (200MB)", () => {
```
HasanCPS, line 8: I’d like us to consolidate the register-case setup so we get the same coverage with less overhead.

At the moment, each of these specs creates a fresh case (`setupTestData -> registerCase`) and repeats Egress + NetApp connect:

- `e2e/pw/tests/egress-to-netapp-copy.spec.ts`
- `e2e/pw/tests/egress-to-netapp-copy-large.spec.ts`
- `e2e/pw/tests/egress-to-netapp-copy-multifile.spec.ts`
- `e2e/pw/tests/netapp-to-egress-copy.spec.ts`
- `e2e/pw/tests/login-only.spec.ts`

(`e2e/pw/tests/egress-to-netapp-move.spec.ts` is currently skipped.)

Can we align on creating and connecting one case once, then reusing that connected case across the register-case specs?
That should keep intent clear while reducing runtime, flakiness, and case/workspace churn.

e2e/pw/tests/egress-to-netapp-copy-default.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+    await transferTab.confirmTransfer();
+
+    // Step 6: Wait for transfer to complete
+    await transferTab.waitForTransferComplete();
+
+    // Save the transferred filename so netapp-to-egress-copy-default can pick it up
+    const sharedFile = path.resolve(__dirname, "../test-results/.last-transferred-file.txt");
+    fs.mkdirSync(path.dirname(sharedFile), { recursive: true });
+    fs.writeFileSync(sharedFile, testData.files[0].fileName);
+
```
HasanCPS: This spec-to-spec dependency via `test-results/.last-transferred-file.txt` is
brittle. It fails if the tests run out of order, in isolation, in parallel
shards, or after `test-results/` is cleaned. The silent fallback to
`testData.files[0].fileName` will succeed against a stale file on NetApp and
hide regressions.

Options:
  1. Make `netapp-to-egress-copy-default` self-sufficient by picking *any*
     recent file from the NetApp panel (e.g. sort by modified-date desc and
     take row 0), which is what the original `netapp-to-egress-copy.spec.ts`
     does.
  2. If coupling is intended, move both assertions into a single spec with
     two `test.step` blocks so ordering and state are explicit.

Please remove the shared file hand-off before merge.

e2e/pw/pages/CaseSearchPage.ts
kmshrajcps: updated

e2e/pw/pages/TransferMaterialsTab.ts
kmshrajcps: updated

e2e/pw/.gitignore
HasanCPS: Where is that? I can’t see it. This is still under e2e/pw/.gitignore. What I’m suggesting is adding it at the root level of the repo here: [https://github.com/CPS-Innovation/Large-and-Complex-Cases/blob/FCT2-14888/.gitignore](https://github.com/CPS-Innovation/Large-and-Complex-Cases/blob/FCT2-14888/.gitignore)

e2e/pw/pages/CaseSearchPage.ts
HasanCPS: I can't see the changes. Which commit is that?

e2e/pw/helpers/egress-api.ts
```diff
@@ -0,0 +1,396 @@
...
+  let maxNumber = 0;
+
+  for (let page = 1; page <= maxPages; page++) {
+    const response = await fetch(
+      `${baseUrl}/api/v1/workspaces/?page=${page}&page_size=${pageSize}`,
+      { headers: { Authorization: `Basic ${token}` } }
+    );
+
+    if (!response.ok) break;
+
```
HasanCPS, line 99: `if (!response.ok) break` still treats workspace-list failures as normal pagination completion. The function can then return `AUTOMATION-TESTING...` at lines 107-109 even after a 401/403/5xx.

Please throw with status + body on non-OK, consistent with `authenticateEgress` and `createWorkspace`.

e2e/pw/helpers/egress-api.ts
```diff
@@ -0,0 +1,396 @@
...
+  const maxPages = 50;
+  const results: EgressWorkspaceInfo[] = [];
+
+  for (let page = 1; page <= maxPages; page++) {
+    const response = await fetch(
+      `${baseUrl}/api/v1/workspaces/?page=${page}&page_size=${pageSize}&view=full`,
+      { headers: { Authorization: `Basic ${token}` } }
+    );
+    if (!response.ok) break;
+
```
HasanCPS, line 56: Uses the same `if (!response.ok) break` pattern at `egress-api.ts` lines 44-50. In teardown (`register-case.teardown.ts` lines 43-47), that means a failed list request looks like "no stale workspaces" rather than "cleanup did not run."

Please throw on non-OK here too, or catch/log explicitly in teardown if cleanup is intentionally best-effort.

e2e/pw/tests/netapp-to-egress-copy.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+    // file we pick won't collide with prior runs' copies.
+    const dateHeader = page
+      .getByTestId("netapp-table-wrapper")
+      .getByRole("button", { name: "Last modified date" });
+    await dateHeader.click();
+    await transferTab.waitForNetAppFiles();
+    await dateHeader.click();
+    await transferTab.waitForNetAppFiles();
+
+    await transferTab.selectNetAppFiles([0]);
```
HasanCPS: The NetApp-source specs sort and select row 0 (`netapp-to-egress-copy.spec.ts` lines 39-50, `netapp-to-egress-move.spec.ts` lines 41-50). The fixture uploads per-test files to Egress, not NetApp, so these tests depend on whatever pre-existing NetApp file happens to be present and sorted first.

Please seed/select a deterministic NetApp source file or document/enforce the required pre-existing fixture data.

e2e/pw/tests/seed-netapp-fixture.setup.ts
```diff
@@ -0,0 +1,114 @@
...
+  await transferTab.waitForEgressFiles();
+
+  await transferTab.waitForEgressFileByName(NETAPP_FIXTURE_FILENAME, [
+    SEED_PARENT,
+    SEED_SUBFOLDER,
+  ]);
+  await transferTab.selectEgressFileByName(NETAPP_FIXTURE_FILENAME);
+  await transferTab.selectAction("Copy");
+  await transferTab.confirmTransfer();
+  await transferTab.waitForTransferComplete();
```
HasanCPS: `README.md` lines 138-142 say re-running `seed-netapp-fixture` while the fixture exists is a no-op. But `seed-netapp-fixture.setup.ts` lines 92-99 always initiates Copy and waits for transfer success. If `lcc-e2e-fixture-source.txt` already exists on NetApp, this should go through the duplicate-file error path and fail.

Please either make the seed genuinely idempotent or update the docs to say it must only be run after removing the existing fixture.

e2e/pw/pages/TransferMaterialsTab.ts
```diff
@@ -0,0 +1,266 @@
...
+   */
+  async selectNetAppFileByExactName(fileName: string) {
+    const row = this.page
+      .getByTestId("netapp-table-wrapper")
+      .locator("tbody tr", { hasText: fileName });
+    if ((await row.count()) === 0) {
+      throw new Error(
+        `Required NetApp fixture '${fileName}' not found in the NetApp panel.\n` +
+          `  Seed it via: npx tsx scripts/seed-netapp-fixture.ts\n` +
+          `  See README "Required NetApp fixture" for details.`
```
HasanCPS, line 148: It tells users to run `npx tsx scripts/seed-netapp-fixture.ts`, but that script does not exist. The seed is implemented as `tests/seed-netapp-fixture.setup.ts` and run via the `seed-netapp-fixture` Playwright project.

Please update the error and `constants.ts` comment to the real command, e.g. `RUN_SEED=1 npx playwright test --project=seed-netapp-fixture` (and ideally include PowerShell syntax for Windows).

e2e/pw/tests/netapp-to-egress-copy.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+    // and uses the seeded fixture path instead.
+    const dateHeader = page
+      .getByTestId("netapp-table-wrapper")
+      .getByRole("button", { name: "Last modified date" });
+    await dateHeader.click();
+    await transferTab.waitForNetAppFiles();
+    await dateHeader.click();
+    await transferTab.waitForNetAppFiles();
+
+    await transferTab.selectNetAppFiles([0]);
```
HasanCPS: Default mode now selects `NETAPP_FIXTURE_FILENAME`, but `netapp-to-egress-copy.spec.ts` lines 39-53 still sorts and selects row 0 in register-case mode. The fresh Egress destination avoids collisions, but the source still depends on whatever happens to exist in the shared `Automation-Testing` NetApp folder.

Please document that fixture requirement clearly or select a deterministic source file here too.

e2e/pw/tests/seed-netapp-fixture.setup.ts
```diff
@@ -0,0 +1,129 @@
...
+    SEED_SUBFOLDER
+  );
+  const uploaded = await uploadFile(
+    config.egressBaseUrl,
+    egressToken,
+    config.defaultWorkspaceId,
+    SEED_SIZE_MB * 1024 * 1024,
+    NETAPP_FIXTURE_FILENAME,
+    `${SEED_PARENT}/${SEED_SUBFOLDER}/`
+  );
```
HasanCPS, line 68: **NetApp fixture seed likely writes the fixture to the wrong path**

`seed-netapp-fixture.setup.ts` lines 61-68 uploads the seed source to Egress under `4. Served Evidence/lcc-e2e-fixture-seed/` and then copies it through the normal Egress-to-NetApp flow. That transfer preserves the source path, so the likely NetApp object is:

`<NETAPP_OPERATION_NAME>/4. Served Evidence/lcc-e2e-fixture-seed/lcc-e2e-fixture-source.txt`

But `netapp-to-egress-copy-default.spec.ts` lines 36-39 selects `lcc-e2e-fixture-source.txt` by exact name from the currently loaded NetApp root. The seed can therefore report success while the default test still fails with "fixture missing".

Please either seed the object at the exact NetApp root path the default test expects, or navigate the default test to the nested folder where the seed actually creates it.

e2e/pw/fixtures/test-fixtures-register-case.ts
```diff
@@ -0,0 +1,162 @@
...
+          config.cmsUsername,
+          config.cmsPassword
+        );
+        for (const file of files) {
+          await deleteNetAppFile(
+            config.lccApiBaseUrl,
+            REGISTER_CASE_NETAPP_FOLDER,
+            file.fileName,
+            accessToken,
+            cmsAuth
```
HasanCPS: **NetApp teardown deletes only the basename, so copied files can be left behind**

The fixture teardown calls `deleteNetAppFile(..., file.fileName)` (`test-fixtures-register-case.ts` lines 149-154; same pattern in the default fixtures). The backend prefixes this with `{operationName}/`, so it attempts to delete:

`<operationName>/<fileName>`

But Egress-to-NetApp transfers from these tests preserve the Egress source path, so the copied object is likely under:

`<operationName>/4. Served Evidence/e2e-.../<fileName>`

That means passing tests can still leave NetApp objects behind. Please pass the full NetApp-relative path to the delete helper, including the upload subfolder/source path for Egress-to-NetApp flows.

e2e/pw/helpers/env-config.ts
```diff
@@ -0,0 +1,76 @@
...
+    e2eAdUser: requireEnv("E2E_AD_USER"),
+    e2eAdPassword: requireEnv("E2E_AD_PASSWORD"),
+    cmsUsername: requireEnv("CMS_USERNAME"),
+    cmsPassword: requireEnv("CMS_PASSWORD"),
+    ddeiAccessKey: requireEnv("DDEI_ACCESS_KEY"),
+    // Separate function key for the case-register endpoint. The
+    // case-register backend rejects the lcc-app DDEI key with HTTP 400
+    // "Unauthorized access to CMS"; only the case-register key is
+    // accepted there. All other DDEI auth flows still use ddeiAccessKey.
+    ddeiAccessKeyCaseRegister: requireEnv("DDEI_ACCESS_KEY_CASE_REGISTER"),
```
HasanCPS: **Default mode should not require the case-register DDEI key**

`loadEnvConfig()` requires `DDEI_ACCESS_KEY_CASE_REGISTER` unconditionally. Default mode calls this loader via `setup-helper-default.ts` line 27 but does not register cases, so `npm run e2e:existing-case` now requires an unrelated register-case secret.

Please make that env var optional in shared config and fail only in the register-case setup path that actually uses it.

e2e/pw/fixtures/test-fixtures-default.ts
```diff
@@ -0,0 +1,68 @@
...
+        config.egressServiceAccountAuth
+      );
+      const fileIds = result.files
+        .map((f) => f.id)
+        .filter((id): id is string => !!id);
+      await deleteFiles(
+        config.egressBaseUrl,
+        token,
+        result.workspace.id,
+        fileIds
```
HasanCPS: **NetApp-to-Egress copied destination files are not cleaned up**

The fixture cleanup deletes the Egress files uploaded during setup (`test-fixtures-default.ts` lines 28-35 and `test-fixtures-register-case.ts` lines 123-129). For NetApp-to-Egress tests, the file copied into `2. Counsel only/<uploadSubfolder>/` is created during the test and is not included in that delete list.

That means passing NetApp-to-Egress tests can leave copied fixture files in the shared/default workspace. Please either delete the copied destination file by path/id after these tests, or update the README to make this cleanup limitation explicit.

ui-spa/src/schemas/responses/searchResult.ts
```diff
@@ -1,13 +1,24 @@
 import { z } from "zod";
 
+const activeManageMaterialsOperationSchema = z.object({
+  id: z.string(),
```
renjithabby: sorry , why we have changes in the schema?

kmshrajcps: this shouldnt be there

e2e/pw/pages/ActivityLogTab.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  async getLatestActivity() {
+    return this.page
+      .getByTestId("activities-timeline")
+      .locator("section")
+      .first();
+  }
+
+  async verifyTransferLogged(transferType: "Copy" | "Move") {
```
HasanCPS: **Activity log assertion can pass against an older transfer**

`ActivityLogTab.verifyTransferLogged()` only verifies the first activity section matching generic Copy/Move text (`ActivityLogTab.ts` lines 23-57). Default mode reuses the same existing case across runs, so older successful Copy/Move entries can remain in the timeline. That means a test can pass the activity-log check even if the current transfer did not create the expected activity entry.

Please pass a current-run value into the assertion, such as the generated file name or `uploadSubfolder`, and assert the matched section contains it before expanding the file list or downloading the CSV.

HasanCPS: @kmshrajcps can we resolve the issue above?

HasanCPS (dismissed):
LGTM!

e2e/pw/helpers/netapp-api.ts
```diff
@@ -0,0 +1,107 @@
...
+): Promise<void> {
+  if (!baseUrl || !operationName) {
+    console.warn(
+      `  [teardown] deleteNetAppFile skipped — missing LCC_API_BASE_URL or NETAPP_OPERATION_NAME`
+    );
+    return;
+  }
+  try {
+    const response = await fetch(
+      `${baseUrl}/api/v1/cases/${encodeURIComponent(operationName)}/netapp`,
```
HasanCPS: Call the supported NetApp delete route during teardown. When `LCC_API_BASE_URL` is configured, passing Egress→NetApp tests call this teardown helper, but the API in this repo exposes
    `POST /api/v1/netapp/delete/batch`, not` DELETE /api/v1/cases/{operationName}/netapp`. Because non-OK responses are only logged, the suite stays green while generated files are never removed from the shared NetApp folders.

e2e/pw/fixtures/test-fixtures-register-case.ts
```diff
@@ -0,0 +1,136 @@
...
+    .replace(/^-|-$/g, "")
+    .slice(0, 40);
+}
+
+export const test = base.extend<
+  { testOptions: RegisterCaseTestOptions; testData: TestSetupResult }
+>({
+  testOptions: [{ fileSizeMb: 100, fileCount: 1 }, { option: true }],
+
+  testData: async ({ testOptions, page }, use, testInfo) => {
```
HasanCPS: Give the register-case fixture its own timeout. For register-case specs with large/multifile options, this fixture uploads up to 200MB or 3x50MB and performs browser login before the test body starts; the `test.setTimeout(...)` calls inside the specs execute only after this fixture has finished, so setup still uses the config's 120s timeout and can fail before the test gets a chance to extend it. The default-mode fixtures already set a fixture timeout, and this fixture needs the same treatment or a timeout declared outside the test body.

e2e/pw/package.json
```diff
@@ -0,0 +1,25 @@
...
+    "e2e:ci": "npx playwright test",
+    "e2e:existing-case": "npx playwright test --project=default-mode-tests",
+    "e2e:existing-case:ui": "npx playwright test --ui --project=default-mode-tests",
+    "e2e:register-case": "npx playwright test --project=register-case-tests",
+    "e2e:register-case:ui": "npx playwright test --ui --project=register-case-tests",
+    "e2e:headed": "npx playwright test --headed",
+    "typecheck": "tsc --noEmit",
+    "report": "npx playwright show-report ./playwright-report",
+    "report:generate": "npx playwright test --reporter=html",
+    "report:junit": "npx playwright test --reporter=junit > test-results/results.xml"
```
HasanCPS: Create the JUnit output directory before redirecting. On a clean checkout `e2e/pw/test-results/` is ignored and absent, and shell redirection opens `test-results/results.xml` before Playwright starts. That makes `npm run report:junit` fail immediately with a missing-directory error instead of producing the documented CI report.

HasanCPS (approved):
LGTM!

PR #344, opened 2026-04-09
add shared infra for manage materials stories

bridgesr:


backend/CPS.ComplexCases.FileTransfer.API.Tests/Unit/Factories/StorageClientFactoryTests.cs
```diff
@@ -93,6 +93,18 @@ public void GetClientsForDirection_CallsGetClientTwice_WhenDirectionIsNetAppToEg
         Assert.Contains("NetAppStorageClient", exception.Message);
     }
 
+    [Fact]
```
HasanCPS, line 101: Consider adding a test that registers two (or one scoped) `NetAppStorageClient` mocks and asserts `GetClientsForDirection(NetAppToNetApp)` returns `(NetApp, NetApp)` without throwing. The current test is useful but only proves the factory targets NetApp when DI is incomplete.

HasanCPS (approved):
LGTM!

PR #353, opened 2026-04-14
Task/fct2 14896 run postman e2e in cicd

lilachdavis:
- Add pipeline to run Postman tests after successful deployment to staging.
- Update cleanup pipeline to clear the Automation-Testing-Move shared drive folder

Updates to scripts to enable running in CICD:
- Use `$env:TMPDIR` when `$env:TEMP` is unavailable, e.g. on Linux
- Use `curl` on Linux, rather than curl.exe.
- Change `Check-Newman`  to automatically install the necessary tools rather than fail the script. Rename to `Install-Newman`, using pwsh approved verb.
- Use Write-Output instead of Write-Host inside `Install-Newman` following Sonarqube recommendations
- Write JUnit report (instead of json) when running in CI
- Hide sensitive data from html reports when running in CI

Variable cleanup:
- Remove `uiUrl` variable - Not needed, the "origin" header is unnecessary
- Rename `lccApiId` to `lccApiClientId` for increased clarity
- Remove `netappPath` variable - not used anywhere, seems to duplicate `netappFolderPath`
- Remove hardcoded values for `netappFolderPath` and `netappMoveFolderPath` from collection. Set default values in `Run-E2E-Tests.ps1` instead.
- Make `caseApiBaseUrl` var only required for the register-case route/
- Remove var `$jsonReport` - never used down the line


e2e/postman/Teardown-EgressTestFiles.ps1
kmshrajcps, line 1: I want to understand how the delete works, there are tests where it uses predefined cases and workspace, so we shouldnt be deleting those workspace

e2e/postman/Run-E2E-Tests.ps1
```diff
@@ -135,7 +135,7 @@ $Config = @{
     CaseApiBaseUrl = if ($env:LCC_CASE_API_BASE_URL) { $env:LCC_CASE_API_BASE_URL } else { "" }
     EgressBaseUrl = if ($env:LCC_EGRESS_BASE_URL) { $env:LCC_EGRESS_BASE_URL } else { "" }
     DdeiBaseUrl   = if ($env:LCC_DDEI_BASE_URL) { $env:LCC_DDEI_BASE_URL } else { "" }
-    LccApiId      = if ($env:LCC_API_ID) { $env:LCC_API_ID } else { "" }
+    lccApiClientId      = if ($env:LCC_API_CLIENT_ID) { $env:LCC_API_CLIENT_ID } else { "" }
```
HasanCPS: The `lccApiClientId` must be added to both `LCCUserJourneyTests.postman_collection.json` and `LCCTestEnvironment.postman_environment.template.json`; otherwise, the tests will fail.

e2e/postman/Teardown-EgressTestFiles.ps1
lilachdavis: That's explained in the readme:

Deletes test materials uploaded to Egress.

```powershell
# Delete files uploaded to a workspace
.\Teardown-EgressTestFiles -DeleteFiles -WorkspaceId <workspace-id> -FileIds id1,id2,id3 -Force

# Delete an entire workspace
.\Teardown-EgressTestFiles -DeleteWorkspace -WorkspaceId <workspace-id> -Force

File: e2e/postman/Run-E2E-Tests.ps1 (hunk 1 above)
Comment (lilachdavis) [reply e2e/postman/Run-E2E-Tests.ps1]: Done

**Review by kmshrajcps — DISMISSED** [review]
LGTM

File: e2e/postman/Run-E2E-Tests.ps1 [hunk 2]
```diff
@@ -235,19 +241,43 @@ function Write-Info {
...
     $newman = Get-Command newman -ErrorAction SilentlyContinue
     if (-not $newman) {
-        Write-Err "Newman is not installed!"
-        Write-Host ""
-        Write-Host "Install Newman with:" -ForegroundColor Yellow
-        Write-Host "  npm install -g newman" -ForegroundColor White
-        Write-Host "  npm install -g newman-reporter-htmlextra" -ForegroundColor White
-        Write-Host ""
-        return $false
+        write-output "Newman not found. Installing..." -ForegroundColor Yellow
```
HasanCPS: `Write-Output` with `-ForegroundColor` throws a parameter-binding error 

`e2e/postman/Run-E2E-Tests.ps1`, lines 255, 262, 273, 275, the `Write-Output ... -ForegroundColor` calls inside `Install-Newman`.

`Write-Output` has no `-ForegroundColor` parameter. Only `Write-Host` does. On a fresh CI agent where Newman is not preinstalled, `Install-Newman` reaches line 255:

```powershell
write-output "Newman not found. Installing..." -ForegroundColor Yellow
```

PowerShell cannot bind `-ForegroundColor`, so it throws "A parameter cannot be found that matches parameter name 'ForegroundColor'". That is a terminating error, and it can abort the function before the `npm install` line runs. The result is that Newman never installs and the whole run fails at the prerequisite check, which is the exact path this code is meant to handle.

Could we switch these four lines to `Write-Host`, the way the rest of the script already does for coloured output?

e2e/postman/Run-E2E-Tests.ps1
```diff
@@ -133,39 +139,39 @@ $Config = @{
...
-    LccApiId      = if ($env:LCC_API_ID) { $env:LCC_API_ID } else { "" }
+    EgressBaseUrl = if ($env:LCC_EGRESS_BASE_URL) { $env:LCC_EGRESS_BASE_URL } else { "" }
+    LccApiClientId      = if ($env:LCC_API_CLIENT_ID) { $env:LCC_API_CLIENT_ID } else { "" }
     LccApiClientSecret = if ($env:LCC_API_CLIENT_SECRET) { $env:LCC_API_CLIENT_SECRET } else { "" }
     DefaultCaseId      = if ($env:LCC_DEFAULT_CASE_ID) { $env:LCC_DEFAULT_CASE_ID } else { "" }
     DefaultCaseUrn     = if ($env:LCC_DEFAULT_CASE_URN) { $env:LCC_DEFAULT_CASE_URN } else { "" }
     DefaultWorkspaceId   = if ($env:LCC_DEFAULT_WORKSPACE_ID) { $env:LCC_DEFAULT_WORKSPACE_ID } else { "" }
-    DefaultWorkspaceName = if ($env:LCC_DEFAULT_WORKSPACE_NAME) { $env:LCC_DEFAULT_WORKSPACE_NAME } else { "" }
+    DefaultWorkspaceName = if ($env:LCC_DEFAULT_WORKSPACE_NAME) { $env:LCC_DEFAULT_WORKSPACE_NAME } else { "ExistingCaseAutomation" }
+    NetappFolderPath = if ($env:LCC_NETAPP_FOLDER_PATH) { $env:LCC_NETAPP_FOLDER_PATH } else { "${defaultNetappFolderName}/" }
```
HasanCPS: ##  Double slash in the default `NetappFolderPath`

`$defaultNetappFolderName` (set on line 125) already ends in a slash, so it is either `"Automation-Testing/"` or `"existingCaseAutomation/"`. Line 151 then wraps it again:

```powershell
NetappFolderPath = if ($env:LCC_NETAPP_FOLDER_PATH) { $env:LCC_NETAPP_FOLDER_PATH } else { "${defaultNetappFolderName}/" }
```

That produces `Automation-Testing//` or `existingCaseAutomation//`. The value feeds `{{netappFolderPath}}`, which the collection uses six times to build paths and as the FileExists guard, so any run that does not set `LCC_NETAPP_FOLDER_PATH` gets a malformed double-slash path. Every local run hits this. CI dodges it only because `api-postman-test.yml` sets `LCC_NETAPP_FOLDER_PATH` explicitly, but the default itself is still wrong.

Could we drop the trailing slash here and use just `$defaultNetappFolderName`, since it already carries one?

e2e/postman/Setup-EgressWorkspaceAndUpload.ps1
```diff
@@ -698,7 +720,9 @@ if ($SkipUpload) {
             if ($i -gt 1) { Start-Sleep -Seconds $retryDelay }
             Write-Host "  Attempt $i/$retryCount..." -NoNewline
 
-            $filesResult = curl.exe --silent --location "$BaseUrl/api/v1/workspaces/$WorkspaceId/files?path=$encodedFolder" `
+            Start-Sleep -Seconds $retryDelay
```
HasanCPS: ##  Duplicated retry sleep in the file-verification loop

Line 720 keeps the existing conditional sleep, `if ($i -gt 1) { Start-Sleep -Seconds $retryDelay }`, which deliberately skips the wait on the first attempt. Line 723 then adds an unconditional `Start-Sleep -Seconds $retryDelay`. The net effect is that attempt 1 now sleeps when it used to start immediately, and every retry sleeps `2 × $retryDelay`. The file-verification loop runs slower than intended on every upload. It looks like a leftover from the refactor.

Could we remove line 723, or fold it into the existing conditional so the timing matches the original intent?

e2e/postman/Run-E2E-Tests.ps1
lilachdavis: Originally changed to write-output as Sonarqube was flagging Write-Host as a Medium impact maintainability issue. I changed back in the next commit, but as you can see, it's being flagged again. 
Unless we're happy to ignore this issue, I can **either** change back to Write-Output and remove the colour **or** change the cmdlet's name (though it will be less representative of its function).

e2e/postman/Run-E2E-Tests.ps1
lilachdavis: Thanks for spotting. Fixed in next commit.

e2e/postman/Setup-EgressWorkspaceAndUpload.ps1
lilachdavis: Thanks for spotting. Removed in next commit.

e2e/postman/Run-E2E-Tests.ps1
HasanCPS: You asked whether to (a) ignore the SonarQube flag, (b) revert to `Write-Output` and drop the colour, or (c) rename the cmdlet.

**Do not revert to `Write-Output`.** It is the wrong tool on two counts:

- It has no `-ForegroundColor` parameter, which is the binding error that was finding last time.
- More importantly, `Write-Output` writes to the *success/output* stream. `Install-Newman` returns `$true`/`$false` (lines 248, 277, 280) and is meant to be consumed as a boolean. Any `Write-Output` inside it merges those strings into the function's return value, so a caller doing `if (Install-Newman)` gets an array of strings plus the bool instead of a clean boolean. That is a functional bug, not just a style one.

`Write-Host` is the correct cmdlet here — these are user-facing console status lines for a test runner, not pipeline data. The SonarQube rule (`PSAvoidUsingWriteHost`) predates PowerShell 5.0, where `Write-Host` was changed to write through the information stream (redirectable with `6>`), so the "output can't be captured" rationale no longer holds for a console script like this.

**Option (c) is effectively already built into the script**, and is the cleanest path:

The script already has `Write-Header`, `Write-Success`, `Write-Err`, and `Write-Info` wrappers (lines 220–241), each of which is the single place that calls `Write-Host` with a colour. The four `Install-Newman` lines are the only spots that bypass those wrappers and call `Write-Host` directly. Route them through the existing helpers:

- line 255 → `Write-Info "Newman not found. Installing..."` (Write-Info is already Yellow)
- line 262 → `Write-Info "newman-reporter-htmlextra not found. Installing..."`
- line 273 → `Write-Success "Newman and required reporters installed successfully."` (Write-Success is already Green)
- line 275 → `Write-Success "Newman and reporters already installed."`

That removes four raw `Write-Host` hits and shrinks the SonarQube surface to just the handful of calls inside the four wrapper functions. Suppress the rule once there with a justified inline suppression (`[Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSAvoidUsingWriteHost', '')]` on the wrappers, or a project-level PSScriptAnalyzer settings exclusion), rather than fighting it call-by-call. That keeps the colour, keeps the semantics correct, and satisfies the linter in one place.

**Recommendation:** route the four lines through `Write-Info`/`Write-Success`, then add one justified suppression on the wrapper functions. Ignoring the flag ad-hoc (option a) also works but leaves the four inconsistent direct calls in place.

lilachdavis: Got it, thank you!

e2e/postman/Setup-EgressWorkspaceAndUpload.ps1
HasanCPS: Can we use `$TempFolder` here instead of `$env:TEMP`?

This PR adds the cross-platform temp fallback (`$env:TEMP` -> `$env:TMPDIR` -> `/tmp`) at the top of the script, and the later temp files (`egress_init/chunk/complete, lines 479/550/666/667`) correctly use `$TempFolder`. But the workspace-create body file here (line 357) and the add-user body file (line 394) still use `$env:TEMP`.

On the Linux agent the new pipeline runs on, `$env:TEMP` is unset, so `Join-Path $env:TEMP ...` fails and new-workspace creation breaks. The scheduled `api-postman-test.yml` run doesn't hit this, since it passes `LCC_DEFAULT_WORKSPACE_ID` and reuses an existing workspace. It only bites the `-RegisterCase` / new-workspace path, which is the same path this cross-platform change was added for.

Suggested fix: change both lines to `Join-Path $TempFolder "egress_create.json"` and `Join-Path $TempFolder "egress_adduser.json"`.

e2e/postman/README.md
HasanCPS: This variable was renamed to `LCC_API_CLIENT_ID` in this PR (`secrets.config.template.ps1, Run-E2E-Tests.ps1, api-postman-test.yml`), and README line 105 already uses the new name. This table row still says `LCC_API_ID`.

Anyone following the table sets `LCC_API_ID`, then `Run-E2E-Tests.ps1` fails validation with "missing LCC_API_CLIENT_ID".

Suggested fix: update the row to `LCC_API_CLIENT_ID`.

devops-pipelines/cleanup-testfiles.yml
```diff
@@ -128,3 +128,18 @@ jobs:
...
+        displayName: "Cleanup Postman Move Test NetApp folder"
+        condition: succeededOrFailed()
+        inputs:
+          filePath: "$(System.DefaultWorkingDirectory)/devops-pipelines/scripts/test-automation-cleanup/Remove-SharedDriveFolderContent.ps1"
+          arguments: >
+            -ClientId $(CallingAppValidAudience)
+            -TenantId $(TenantId)
+            -AadUsername $(AadUserName)
+            -AadPassword $(AadUserPassword)
+            -CaseId $(MoveTestsCaseId)
```
HasanCPS, line 142: The new "Cleanup Postman Move Test NetApp folder" step passes `-CaseId $(MoveTestsCaseId)`, but the scheduled Postman run never touches that case. `api-postman-test.yml:66` sets `LCC_DEFAULT_CASE_ID: $(E2eTestsCaseId)`, and the move tests in the collection ("E2E: Egress to NetApp Move", "E2E: Netapp Move to Egress Copy") act on `{{caseId}}`, which the collection sets from `defaultCaseId`. So the move test writes its NetApp files into `Automation-Testing-Move/` under `E2eTestsCaseId`, while this step clears that folder under `MoveTestsCaseId`.

If the two variables point at different cases, and the separate name suggests they do, the move test's files are never cleaned and build up on every nightly run, while the step wipes a folder nothing writes to. A folder that keeps filling can then break the move test's `FileExists` prerequisite on later runs. The step directly above it, "Cleanup existingCaseAutomation NetApp folder" (line 126), uses `-CaseId $(E2eTestsCaseId)` for the same run, which is what makes this one look off.

**Suggested Fix:** change line 142 to `-CaseId $(E2eTestsCaseId)`. If the move test is meant to run against `MoveTestsCaseId`, pass that case into `api-postman-test.yml` as `LCC_DEFAULT_CASE_ID` instead, so the test and its cleanup agree on one case.

lilachdavis: The default case ID is associated with the egress folder `ExistingCaseAutomation` and the netapp folder `existingCaseAutomation` which all the **copy** tests use for both source and destination.

According to Kamesh, the **move** test has to use a different netapp folder to avoid conflicts. 
So I'd need to associate the same case ID with two separate netapp folders, which I don't think is possible?

For that reason, I created a new case in CMS and associated it with the Move netapp folder - you can see it getting cleaned up successfully in [this pipeline test run](https://dev.azure.com/CPSDTS/Information%20Management/_build/results?buildId=170372&view=logs&j=77a72ac3-3b77-5ad7-8b77-83ad59f05347&t=6258527a-8a01-59d2-6c56-66bec79452e7)

P.S. Not in the scope of this ticket, there's another cleanup issue with the register case route, since the `Automation-Testing/` netapp folder gets associated with a new case ID every time. Unlike the pw tests, here Kamesh hasn't implemented automated cleanup of the netapp folder, and I don't have a persisting case ID associated with it, which means I haven't been able to add it to the nightly cleanup.

HasanCPS (approved):
LGTM!

PR #355, opened 2026-04-15
add diff check as condition to run database migration job

lilachdavis:


devops-pipelines/backend/backend-build-and-deploy.yml
HasanCPS, line 1: This PR replaces the old backend build/deploy pipelines with `backend-build-and-deploy.yml` but `README.md` still points to the `backend-build-and-publish.yml` and `backend-deploy.yml` (both deleted). Can we update the backend pipeline docs in README in the same PR so onboarding/runbook links stay correct?

lilachdavis: README updated

HasanCPS (approved):
LGTM!

PR #356, opened 2026-04-17
Feature/fct2 10846 delete materials

bridgesr:


backend/CPS.ComplexCases.NetApp/Factories/NetAppArgFactory.cs
```diff
@@ -157,14 +157,15 @@ public RegenerateUserKeysArg CreateRegenerateUserKeysArg(string username, string
...
+    public DeleteFileOrFolderArg CreateDeleteFileOrFolderArg(string bearerToken, string bucketName, string operationName, string path, bool isFolder = false)
     {
         return new DeleteFileOrFolderArg
         {
             BearerToken = bearerToken,
             BucketName = bucketName.ToLowerInvariant(),
             OperationName = operationName,
-            Path = path
+            Path = path,
+            IsFolder = isFolder
```
HasanCPS, line 168: `DeleteNetAppFileOrFolder`: `CreateDeleteFileOrFolderArg` now requires an accurate **`isFolder`** for recursive folder delete in `NetAppClient`; this call omits it (**73–74**), so folder deletes only hit a single key.
**Preferred:** remove this function and route (403 in Production anyway) and standardize on **`POST v1/netapp/delete/batch`** so we do not maintain two delete contracts. Only invest in DTO/`isFolder`/tests here if removal is genuinely blocked by an external caller.

backend/CPS.ComplexCases.API/Functions/DeleteNetAppBatch.cs
```diff
@@ -0,0 +1,155 @@
...
+        var bucket = securityGroups.First().BucketName;
+
+        var results = new List<DeleteNetAppBatchItemResult>();
+
+        foreach (var op in batchRequest.Value.Operations)
+        {
+            var isFolder = op.Type == NetAppDeleteOperationType.Folder;
+            var arg = _netAppArgFactory.CreateDeleteFileOrFolderArg(
+                context.BearerToken, bucket, string.Empty, op.SourcePath, isFolder);
+
```
HasanCPS, line 98: **Batch delete: `sourcePath` is not scoped to the case (unlike search); can return success without deleting**

`DeleteNetAppBatch` passes `CreateDeleteFileOrFolderArg(..., operationName: string.Empty, path: op.SourcePath, ...)`. `NetAppRequestFactory.DeleteObjectRequest` sets S3 `Key = arg.Path` with no case prefix applied in the batch function.

Search resolves the case’s `NetappFolderPath` and uses it as the NetApp scope; batch does not. If callers send paths as shown in the UI (relative to the case root) **without** the same bucket key prefix the objects actually use, `DeleteObject` targets a **non-existent key**. On S3-compatible APIs, deleting a missing key is **idempotent and still succeeds**, so the API can return **200 OK** with per-item success while **nothing visible changes** in NetApp.

Please load case metadata for `caseId` (same family as search), require non-empty `NetappFolderPath`, normalize paths, and either (a) require each `sourcePath` to start with that prefix, or (b) join `NetappFolderPath` + relative path if the contract is “paths relative to case root.” Without this, `caseId` also does not **authorize** the S3 keys being deleted (path traversal rules alone are insufficient).

backend/CPS.ComplexCases.API/Validators/Requests/DeleteNetAppBatchRequestValidator.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    public DeleteNetAppBatchRequestValidator()
+    {
+        RuleFor(x => x.Operations)
+            .NotEmpty()
+            .WithMessage("Operations cannot be empty.");
+
+        RuleFor(x => x.Operations)
+            .Must(ops => ops == null || ops.Select(o => o.SourcePath).Distinct(StringComparer.OrdinalIgnoreCase).Count() == ops.Count)
+            .WithMessage("Duplicate sourcePath values are not permitted in a single batch.");
```
HasanCPS, line 26: Consider capping `Operations.Count` (FluentValidation `Must` with a reasonable maximum). An unbounded list can create very long-running function invocations, large responses, and unnecessary load on S3/NetApp. Align the limit with product expectations and any gateway timeout.

backend/CPS.ComplexCases.API/Functions/DeleteNetAppBatch.cs
```diff
@@ -0,0 +1,155 @@
...
+    [CmsAuthValuesAuth]
+    [BearerTokenAuth]
+    [OpenApiRequestBody(ContentType.ApplicationJson, typeof(DeleteNetAppBatchDto), Description = "Body containing the case ID and list of files/folders to delete.")]
+    [OpenApiResponseWithBody(statusCode: HttpStatusCode.OK, contentType: ContentType.ApplicationJson, bodyType: typeof(DeleteNetAppBatchResponse), Description = ApiResponseDescriptions.Success)]
+    [OpenApiResponseWithBody(statusCode: HttpStatusCode.BadRequest, contentType: ContentType.TextPlain, typeof(string), Description = ApiResponseDescriptions.BadRequest)]
+    [OpenApiResponseWithBody(statusCode: HttpStatusCode.Unauthorized, contentType: ContentType.TextPlain, typeof(string), Description = ApiResponseDescriptions.Unauthorized)]
+    [OpenApiResponseWithBody(statusCode: HttpStatusCode.Forbidden, contentType: ContentType.TextPlain, typeof(string), Description = ApiResponseDescriptions.Forbidden)]
+    [OpenApiResponseWithBody(statusCode: HttpStatusCode.InternalServerError, contentType: ContentType.TextPlain, typeof(string), Description = ApiResponseDescriptions.InternalServerError)]
```
HasanCPS: **`DeleteNetAppBatch`:** BadRequest is declared as `TextPlain` / `string`, but the code returns `BadRequestObjectResult(batchRequest.ValidationErrors)` (**58–59**). Align content type and `bodyType` with the real payload.

backend/CPS.ComplexCases.API/Validators/Requests/DeleteNetAppBatchRequestValidator.cs
```diff
@@ -0,0 +1,33 @@
+using CPS.ComplexCases.Data.Models.Requests;
+using FluentValidation;
+
+namespace CPS.ComplexCases.API.Validators.Requests;
+
+public class DeleteNetAppBatchRequestValidator : AbstractValidator<DeleteNetAppBatchDto>
+{
+    public DeleteNetAppBatchRequestValidator()
```
HasanCPS, line 10: `CaseId` is used for activity logging but is not validated (e.g. `> 0`). Invalid IDs may produce confusing audit rows.

backend/CPS.ComplexCases.API/Functions/DeleteNetAppBatch.cs
```diff
@@ -0,0 +1,155 @@
...
+                            ? ActivityLog.Enums.ResourceType.NetAppFolder
+                            : ActivityLog.Enums.ResourceType.Material;
+
+                        await _activityLogService.CreateActivityLogAsync(
+                            actionType,
+                            resourceType,
+                            batchRequest.Value.CaseId,
+                            op.SourcePath,
+                            op.SourcePath,
+                            context.Username);
```
HasanCPS: Today each successful delete writes a separate `CreateActivityLogAsync` (material vs folder). Transfers instead tend to log **once** with a **`JsonDocument` details** payload (`FileTransferDetails`). Consider whether batch delete should mirror that (e.g. one action type + structured details listing each `sourcePath` and outcome) for fewer rows and parity with transfer UX—or keep per-item logs if the activity timeline must show one row per object. Either is defensible; worth an explicit product decision.

backend/CPS.ComplexCases.NetApp/Client/NetAppClient.cs
```diff
@@ -477,58 +477,62 @@ public async Task<bool> DoesObjectExistAsync(GetObjectArg arg)
...
         var s3Client = await _s3ClientFactory.GetS3ClientAsync(arg.BearerToken);
         try
         {
-            if (Path.HasExtension(arg.Path))
+            if (!arg.IsFolder)
             {
                 var request = _netAppRequestFactory.DeleteObjectRequest(arg);
                 await s3Client.DeleteObjectAsync(request);
-                return $"Successfully deleted file {arg.Path} from bucket {arg.BucketName}.";
+                return new DeleteNetAppResult(true, 1, null, null);
```
HasanCPS: After `DeleteObjectAsync`, the code returns `new DeleteNetAppResult(true, 1, null, null)` without checking whether the key existed. That matches S3 idempotency but is **misleading** for APIs and UIs inferring “one object removed.” If you need stricter semantics, consider `HeadObject` first or surface delete response metadata where the provider exposes it.

backend/CPS.ComplexCases.API/Functions/DeleteNetAppBatch.cs
bridgesr: @HasanCPS all other changes are made, this one is for the product team to decide

backend/CPS.ComplexCases.API/Functions/DeleteNetAppBatch.cs
```diff
@@ -0,0 +1,189 @@
...
+                }.SerializeToJsonDocument(_logger);
+
+                await _activityLogService.CreateActivityLogAsync(
+                    ActivityLog.Enums.ActionType.MaterialDeleted,
+                    ActivityLog.Enums.ResourceType.Material,
+                    batchRequest.Value.CaseId,
+                    batchRequest.Value.CaseId.ToString(),
+                    null,
+                    context.Username,
+                    details);
```
HasanCPS, line 202: `DeleteNetAppBatch` now logs `details` as `{ items: [...] }` for `ActionType.MaterialDeleted`. The current Activity Log UI treats any non-null `activity.details` as **transfer** details and immediately dereferences `details.sourcePath` / `details.destinationPath` / `details.files` / `details.errors` (`ActivityTimeline.tsx`).
This is a contract mismatch and can break rendering when non-transfer entries include a different details shape.

Please update the UI to branch by `actionType` and parse typed details safely before dereferencing fields.
Also consider using `MaterialOrFolderDeleted` when all operations are folders, or a dedicated batch-delete action type if mixed.

backend/CPS.ComplexCases.NetApp/Client/NetAppClient.cs
HasanCPS, line 235: When `ListObjectsInBucketAsync` returns `null` (**232–235**), `ListAllObjectKeysForDeletionAsync` (**877–923**) still ends up adding only `prefix` (**920**) and can proceed with a partial key set. Please **fail the folder delete** (or retry) when `listResponse` is null so we never report success with children still under the prefix.

backend/CPS.ComplexCases.NetApp/Client/NetAppClient.cs
```diff
@@ -477,58 +477,66 @@ public async Task<bool> DoesObjectExistAsync(GetObjectArg arg)
...
+            if (!arg.IsFolder)
             {
+                var getObjectArg = _netAppArgFactory.CreateGetObjectArg(arg.BearerToken, arg.BucketName, arg.Path);
+                if (!await DoesObjectExistAsync(getObjectArg))
+                    return new DeleteNetAppResult(true, 0, null, null);
+
                 var request = _netAppRequestFactory.DeleteObjectRequest(arg);
                 await s3Client.DeleteObjectAsync(request);
-                return $"Successfully deleted file {arg.Path} from bucket {arg.BucketName}.";
+                return new DeleteNetAppResult(true, 1, null, null);
```
HasanCPS: For missing files, `NetAppClient` returns success with `KeysDeleted = 0` (**488–496**), and the batch API still labels the row **`Deleted`** (**116–123**). Consider a distinct outcome (e.g. `NotFound`) or document idempotent semantics in the API contract.

backend/CPS.ComplexCases.API/Functions/DeleteNetAppBatch.cs
```diff
@@ -0,0 +1,219 @@
...
+                }
+                else
+                {
+                    results.Add(new DeleteNetAppBatchItemResult
+                    {
+                        SourcePath = op.SourcePath,
+                        Status = "Deleted",
+                        KeysDeleted = result.KeysDeleted > 1 ? result.KeysDeleted : null
+                    });
+                }
```
HasanCPS, line 133: For folders, we do not verify existence before delete like files do. A missing folder can still flow through as success with `WasFound=true` and become `Status="Deleted"` in API response, which may be misleading.

For folder deletes, `DeleteFileOrFolderCoreAsync` currently returns `WasFound=true` on success (`NetAppClient.cs` line 539) and always includes the prefix key (`919-921`), so a non-existent folder can still surface as `Deleted` in batch response.

backend/CPS.ComplexCases.API/Functions/DeleteNetAppBatch.cs
```diff
@@ -0,0 +1,219 @@
...
+        }
+
+        return new OkObjectResult(new DeleteNetAppBatchResponse
+        {
+            TotalRequested = batchRequest.Value.Operations.Count,
+            Succeeded = succeeded,
+            NotFound = notFound,
+            Failed = failed,
+            Results = results
+        });
```
HasanCPS, line 226: Consider adding a root `Status` field so clients do not re-derive outcome from counters every time. Suggested values: `Completed`, `PartiallyCompleted`, `Failed` (optionally `NoOp` when all items are `NotFound`). You already compute `Succeeded`/`NotFound`/`Failed` (lines 157-160), so this is straightforward to add at response assembly.

backend/CPS.ComplexCases.API/Functions/DeleteNetAppBatch.cs
```diff
@@ -0,0 +1,219 @@
...
+
+                var hasFolder = auditedOps.Any(op => op.Type == NetAppDeleteOperationType.Folder);
+                var hasMaterial = auditedOps.Any(op => op.Type == NetAppDeleteOperationType.Material);
+
+                var actionType = (hasFolder, hasMaterial) switch
+                {
+                    (true, true) => ActivityLog.Enums.ActionType.FolderAndMaterialDeleted,
+                    (true, false) => ActivityLog.Enums.ActionType.FolderDeleted,
+                    _ => ActivityLog.Enums.ActionType.MaterialDeleted
+                };
```
HasanCPS: It picks action type dynamically (`FolderDeleted` / `MaterialDeleted` / `FolderAndMaterialDeleted` at lines 177-182), but logs always use `ResourceType.Material` (lines 195-198). For folder-only and mixed batches this can skew activity filters/reporting. Please map resource type consistently (e.g. folder-only -> `NetAppFolder`, material-only -> `Material`; mixed -> agreed convention).

backend/CPS.ComplexCases.API/Functions/DeleteNetAppBatch.cs
```diff
@@ -0,0 +1,219 @@
...
+
+        var batchRequest = await _requestValidator.GetJsonBody<DeleteNetAppBatchDto, DeleteNetAppBatchRequestValidator>(req);
+
+        if (!batchRequest.IsValid)
+            return new BadRequestObjectResult(batchRequest.ValidationErrors);
+
+        var caseMetadata = await _caseMetadataService.GetCaseMetadataForCaseIdAsync(batchRequest.Value.CaseId);
+
+        if (caseMetadata == null || string.IsNullOrEmpty(caseMetadata.NetappFolderPath))
+            return new BadRequestObjectResult("Case metadata or NetApp folder path is missing.");
```
HasanCPS: OpenAPI declares 400 as `IEnumerable<string>` (line 49), but the case-metadata guard returns a plain string `BadRequestObjectResult` (lines 67-68). Please align runtime and contract (either always return array/object payloads or update OpenAPI to include this string case) so generated clients are accurate.

backend/CPS.ComplexCases.ActivityLog/Services/ActivityLogService.cs
```diff
@@ -277,6 +277,9 @@ private static string SetDescription(ActionType actionType, string? resourceName
             ActionType.TransferInitiated => $"Transfer initiated between {resourceName}",
             ActionType.TransferCompleted => $"Transfer completed between {resourceName}",
             ActionType.TransferFailed => $"Transfer failed between {resourceName}",
+            ActionType.FolderDeleted => "Folder(s) deleted",
+            ActionType.MaterialDeleted => "Document(s) deleted",
```
HasanCPS: I suggest we keep the terminology consistent. Since we’re already using Material to refer to documents and files, we should stick with that for consistency.

backend/CPS.ComplexCases.ActivityLog/Services/ActivityLogService.cs
```diff
@@ -277,6 +277,9 @@ private static string SetDescription(ActionType actionType, string? resourceName
             ActionType.TransferInitiated => $"Transfer initiated between {resourceName}",
             ActionType.TransferCompleted => $"Transfer completed between {resourceName}",
             ActionType.TransferFailed => $"Transfer failed between {resourceName}",
+            ActionType.FolderDeleted => "Folder(s) deleted",
+            ActionType.MaterialDeleted => "Material(s) deleted",
+            ActionType.FolderAndMaterialDeleted => "Folder(s) and document(s) deleted",
```
HasanCPS: Can we also update terminology for the Material here as well?

backend/CPS.ComplexCases.NetApp/Client/NetAppClient.cs
```diff
@@ -477,58 +477,71 @@ public async Task<bool> DoesObjectExistAsync(GetObjectArg arg)
...
+                    return new DeleteNetAppResult(false, true, totalDeleted,
+                        $"Deletion failed for {allErrors.Count} object(s) in folder {folderPath}.", null);
                 }
 
-                var successfulDeletionsCount = deletedObjects.Count;
-                var failedDeletionsCount = deleteErrors.Count;
-
-                return
-                    $"Successfully deleted {successfulDeletionsCount} files from bucket {arg.BucketName}. Deletion failed for {failedDeletionsCount} files. ";
+                return new DeleteNetAppResult(true, true, totalDeleted, null, null);
```
HasanCPS, line 557: There is still one correctness gap in folder delete semantics:

For files, missing objects are surfaced as `WasFound = false` and API status `NotFound`, but folder deletes still always return `WasFound = true` on the success path, which maps to `Status = "Deleted"` in batch response.

- Folder flow always appends prefix key and returns success/found:
  - `backend/CPS.ComplexCases.NetApp/Client/NetAppClient.cs` lines 499-505, 924-926, 544-545
- Batch mapping:
  - `backend/CPS.ComplexCases.API/Functions/DeleteNetAppBatch.cs` lines 116-133

This means deleting a non-existent folder can still appear as `Deleted` (instead of `NotFound`), unlike file behavior.

Could we align folder behavior with file behavior by introducing a reliable `WasFound` signal for folders?

HasanCPS (approved):
LGM!

PR #360, opened 2026-04-27
fix: empty folder deletion

bridgesr:


backend/CPS.ComplexCases.NetApp/Client/NetAppClient.cs
```diff
@@ -939,6 +940,32 @@ private async Task<IEnumerable<string>> ListAllObjectKeysForDeletionAsync(string
         return objectKeys;
     }
 
+    // Checks whether a folder exists by looking for it in its parent's common-prefix
+    // listing. This is more reliable than a HEAD probe for empty folders that were
+    // created outside the S3 API (e.g. via SMB/NFS) and therefore have no zero-byte
+    // marker key in the bucket.
+    private async Task<bool> DoesFolderExistInParentListingAsync(string bucketName, string folderPath, string bearerToken)
```
HasanCPS, line 952: I like the direction of this fix (fallback from marker HEAD to parent-prefix listing), but I think we still have a correctness gap for empty folders in large parents.

## Scenario to consider

- Deleting `cases/123/Uploads/OldEvidence/` (empty folder, no marker key)
- `HEAD` on folder marker returns `404` (expected for SMB/NFS-created folders)
- Parent `cases/123/Uploads/` has many children, so listing is paginated
- `OldEvidence/` appears on page 2+ (not page 1)

## Current behavior risk

- `DoesFolderExistInParentListingAsync` checks only one page
- If folder is not on first page, we return `WasFound = false` and skip delete
- This can reintroduce the same “empty folder not deleted” behavior for larger directories

## Additional risk

- If parent listing fails transiently (`null` response), we currently treat it as “not found”
- This can mask operational failures as a normal no-op delete

## Suggested changes

1. Iterate parent listing across continuation tokens until match found or pages exhausted
2. Handle listing failure separately from “folder does not exist” (avoid false not-found)
3. Add tests for:
   - `HEAD` 404 + folder present on later parent listing page
   - Parent listing failure path (ensure not silently mapped to not-found)

HasanCPS (approved):
LGTM!

PR #415, opened 2026-04-28
Feature/fct2 10847 batch copy

bridgesr:


backend/CPS.ComplexCases.API/Functions/InitiateBatchCopy.cs
```diff
@@ -0,0 +1,111 @@
...
+            _logger.LogWarning("Case metadata or NetApp folder path missing for CaseId: {CaseId}. CorrelationId: {CorrelationId}",
+                batchRequest.Value.CaseId, context.CorrelationId);
+            return new BadRequestObjectResult(new [] { "Case metadata or NetApp folder path is missing." });
+        }
+
+        var casePrefix = caseMetadata.NetappFolderPath.EndsWith('/')
+            ? caseMetadata.NetappFolderPath
+            : caseMetadata.NetappFolderPath + "/";
+
+        var invalidPaths = batchRequest.Value.Operations
```
HasanCPS, line 79: The source paths are constrained to the case NetApp folder, but `destinationPrefix` is not checked against the same `casePrefix` before being forwarded to FileTransfer. Because the bucket is shared, a caller could copy valid source files from this case into another case's subtree.

**Example:**

- Case `123` has NetApp folder `Cases/123/`
- Request source path: `Cases/123/Evidence/report.pdf`
- Request destination prefix: `Cases/999/Evidence/`

Current behavior:

- `sourcePath` passes validation because it is inside `Cases/123/`
- `destinationPrefix` is forwarded unchanged
- the copy can write into `Cases/999/Evidence/report.pdf`

Expected behavior:

- reject the request because `destinationPrefix` is outside `Cases/123/`

**Suggestion:** after `casePrefix` is computed, validate `batchRequest.Value.DestinationPrefix.StartsWith(casePrefix, StringComparison.OrdinalIgnoreCase)`. Return `BadRequest` if it is outside the case folder, and add a unit test for an out-of-case destination prefix.

backend/CPS.ComplexCases.Data/Repositories/CaseActiveManageMaterialsRepository.cs
```diff
@@ -0,0 +1,88 @@
...
+            foreach (var incomingSource in incomingSourceList)
+            {
+                if (existingSourcePaths.Any(existing => PathsOverlap(incomingSource, existing)))
+                    return true;
+
+                if (existingDestPaths.Any(existing => PathsOverlap(incomingSource, existing)))
+                    return true;
+            }
+
+            // Or if any incoming destination overlaps with an existing source path
```
HasanCPS: This conflict check compares incoming destinations only against existing source paths. It does not compare `incomingDestList` with `existingDestPaths`, so two active copy operations with different sources can still target the same destination prefix/tree. That allows overlapping writes despite the active-operation lock.

**Example:**

Existing active operation:

- `sourcePaths`: `["Cases/123/FolderA/file1.pdf"]`
- `destinationPaths`: `["Cases/123/SharedDest/"]`

Incoming operation:

- `sourcePaths`: `["Cases/123/FolderB/file2.pdf"]`
- `destinationPaths`: `["Cases/123/SharedDest/"]`

Current behavior:

- incoming destination is checked against existing source paths only
- no conflict is found
- both operations can write to `Cases/123/SharedDest/`

Expected behavior:

- destination `Cases/123/SharedDest/` overlaps the existing destination
- return `Conflict`

**Suggestion:** add an incoming destination vs existing destination overlap check using `PathsOverlap`, and add repository tests for source/source, source/destination, destination/source, and destination/destination overlaps.

backend/CPS.ComplexCases.FileTransfer.API/Functions/InitiateBatchCopy.cs
```diff
@@ -0,0 +1,272 @@
...
+                {
+                    var listArg = _netAppArgFactory.CreateListObjectsInBucketArg(
+                        request.BearerToken, request.BucketName,
+                        continuationToken: continuationToken,
+                        prefix: sourcePrefix);
+
+                    var listResult = await _netAppClient.ListObjectsInBucketAsync(listArg);
+                    if (listResult == null) break;
+
+                    foreach (var file in listResult.Data.FileData)
```
HasanCPS, line 130: Folder copy expansion enqueues every source object without checking whether the computed destination key already exists. Material copies do a destination `DoesObjectExistAsync` preflight, but folder copies do not, and `TransferFile` only performs an existing-file check for `EgressToNetApp`. `NetAppToNetApp` folder copies can therefore overwrite existing destination objects.

**Example:**

Request:

- copy folder: `Cases/123/SourceFolder/`
- destination prefix: `Cases/123/Destination/`

Source folder contains:

- `Cases/123/SourceFolder/a.pdf`
- `Cases/123/SourceFolder/b.pdf`

Destination already contains:

- `Cases/123/Destination/SourceFolder/a.pdf`

Current behavior:

- source folder is expanded
- no destination existence check is performed for the expanded files
- `a.pdf` can be overwritten

Expected behavior:

- detect the existing destination key before scheduling
- return `Conflict` and do not start the orchestration

**Suggestion:** during folder expansion, compute the destination key for each listed object and perform the same existence/case-insensitive clash checks as material copies, or list the destination folder prefix once and fail the request if any expanded key would collide.

backend/CPS.ComplexCases.FileTransfer.API/Functions/CleanupStaleManageMaterialsOperations.cs
```diff
@@ -0,0 +1,84 @@
...
+            _config.MaxAgeHours);
+
+        var operations = await _caseActiveManageMaterialsService.GetAllActiveOperationsAsync();
+        var deletedCount = 0;
+
+        foreach (var operation in operations)
+        {
+            try
+            {
+                // Age-based safety net: delete without consulting durable
```
HasanCPS: The age-based cleanup branch deletes active manage-material rows without checking durable orchestration status. A legitimate long-running copy that exceeds `MaxAgeHours` would lose its lock row while still running, allowing another conflicting manage-materials operation to start.

**Example:**

Config:

- `MaxAgeHours = 24`

Operation:

- batch copy of a very large folder starts Monday 01:00
- orchestration is still `Running` Tuesday 02:00

Current behavior:

- cleanup sees the row is older than 24 hours
- row is deleted without checking durable status
- a second copy/move/delete operation can start on overlapping paths

Expected behavior:

- cleanup checks durable status first
- because the orchestration is `Running`, the lock row is retained

**Suggestion:** even for aged rows, query the durable instance and only delete when it is terminal or definitely missing after a safe retry/backoff policy. If a hard max age is intentional, set it above worst-case copy duration and add monitoring/alerting before deleting the row.

backend/CPS.ComplexCases.FileTransfer.API/Functions/InitiateBatchCopy.cs
```diff
@@ -0,0 +1,272 @@
...
+                    "Active transfer blocking batch copy for CaseId: {CaseId}, TransferId: {TransferId}. CorrelationId: {CorrelationId}",
+                    request.CaseId, caseMetadata.ActiveTransferId.Value, correlationId);
+                return new ConflictObjectResult("A case-wide file transfer is in progress. Please wait for it to complete before starting a copy operation.");
+            }
+        }
+
+        // Check for conflicting manage materials operations by path
+        var sourcePaths = request.Operations.Select(op => op.SourcePath).ToList();
+        var destinationPaths = new List<string> { request.DestinationPrefix };
+        var hasConflict = await _caseActiveManageMaterialsService.HasConflictingOperationAsync(
```
HasanCPS: The conflict check happens before NetApp preflight and well before `InsertOperationAsync`. Two concurrent requests can both observe no conflict, both complete preflight, and then both insert active rows and schedule orchestrations. This weakens the active-operation lock under concurrency.

**Example:**

Two users submit copy requests at the same time.

Request A:

- source path: `Cases/123/FolderA/file.pdf`
- destination prefix: `Cases/123/Dest/`

Request B:

- source path: `Cases/123/FolderB/file.pdf`
- destination prefix: `Cases/123/Dest/`

Current behavior:

- A checks conflicts and sees none
- B checks conflicts before A inserts the lock row and also sees none
- both requests insert rows and schedule orchestrations

Expected behavior:

- the conflict check and insert are one atomic operation
- only one request succeeds; the other gets `Conflict`

**Suggestion:** move the final conflict check adjacent to `InsertOperationAsync` and enforce it atomically, for example with a serializable transaction, explicit DB lock, or repository method that checks and inserts in one unit of work.

backend/CPS.ComplexCases.FileTransfer.API/Durable/Activity/WriteCopyActivityLog.cs
```diff
@@ -0,0 +1,83 @@
...
+    [Function(nameof(WriteCopyActivityLog))]
+    public async Task Run([ActivityTrigger] WriteCopyActivityLogPayload payload)
+    {
+        _initializationHandler.Initialize(payload.UserName!, payload.CorrelationId);
+
+        var successfulKeySet = new HashSet<string>(payload.SuccessfulSourceKeys, StringComparer.OrdinalIgnoreCase);
+
+        var items = payload.OriginalOperations.Select(op =>
+        {
+            var isCopied = string.Equals(op.Type, "Folder", StringComparison.OrdinalIgnoreCase)
```
HasanCPS: For folder operations, any successful key under the folder marks the whole folder operation as `Copied`. If a folder contains multiple files and only one succeeds, the activity log reports the folder as copied instead of partial/failed, which can mislead users and auditors.

**Example:**

Original operation:

- copy folder `Cases/123/FolderA/`

Expanded source keys:

- `Cases/123/FolderA/a.pdf`
- `Cases/123/FolderA/b.pdf`
- `Cases/123/FolderA/c.pdf`

Transfer result:

- `a.pdf` succeeds
- `b.pdf` fails
- `c.pdf` fails

Current behavior:

- `successfulKeySet` contains one key under `FolderA/`
- the original folder operation is logged as `Copied`

Expected behavior:

- the log shows `Partial` or `NotCopied` for the folder
- users are not told the full folder copied successfully

**Suggestion:** pass enough expanded folder metadata into `WriteCopyActivityLog` to compare all expected source keys for each folder against successful keys. Then log `Copied` only when all expected keys succeeded, or add an explicit `Partial` outcome.

backend/CPS.ComplexCases.API.Tests/Unit/Functions/InitiateBatchCopyTests.cs
```diff
@@ -0,0 +1,202 @@
...
+        _fileTransferClientMock.Verify(c => c.InitiateBatchCopyAsync(
+            It.Is<CopyNetAppBatchRequest>(r => r.BucketName == _testBucketName),
+            It.IsAny<Guid>()), Times.Once);
+    }
+
+    [Fact]
+    public async Task Run_WithValidRequest_MapsOperationsCorrectly()
+    {
+        var dto = new CopyNetAppBatchDto
+        {
```
HasanCPS, line 222: Current tests cover source paths outside the case folder, but not destination paths.

**Example test case:**

Arrange:

- case NetApp folder path is `Cases/123/`
- `sourcePath` is `Cases/123/file.pdf`
- `destinationPrefix` is `Cases/999/`

Assert:

- function returns `BadRequest`
- `FileTransferClient.InitiateBatchCopyAsync` is not called

backend/CPS.ComplexCases.Data/Repositories/CaseActiveManageMaterialsRepository.cs
```diff
@@ -0,0 +1,88 @@
...
+            .Where(x => x.CaseId == caseId)
+            .ToListAsync();
+    }
+
+    public async Task<IEnumerable<CaseActiveManageMaterialsOperation>> GetAllActiveOperationsAsync()
+    {
+        return await _dbContext.CaseActiveManageMaterialsOperations.ToListAsync();
+    }
+
+    public async Task<bool> HasConflictingOperationAsync(int caseId, IEnumerable<string> sourcePaths, IEnumerable<string> destinationPaths)
```
HasanCPS: This is important locking behavior, especially destination-vs-destination overlap and slash/case normalization cases.

**Example test case:**

Arrange:

- existing operation `destinationPaths` contains `Cases/123/Dest/`
- incoming `destinationPaths` contains `Cases/123/Dest/SubFolder/`
- incoming `sourcePaths` do not overlap existing `sourcePaths`

Assert:

- `HasConflictingOperationAsync` returns `true`

backend/CPS.ComplexCases.FileTransfer.API.Tests/Unit/Functions/InitiateBatchCopyTests.cs
```diff
@@ -0,0 +1,655 @@
...
+
+    [Fact]
+    public async Task Run_WithValidMaterialRequest_OrchestrationInstanceIdMatchesResponseTransferId()
+    {
+        SetupValidRequest();
+        SetupNoActiveTransfer();
+        SetupMaterialSourceExists();
+        SetupMaterialDestinationMissing();
+        SetupNoClashAtDestination();
+        var client = CreateSchedulingClient(out var capturedCalls);
```
HasanCPS, line 425: This should assert that the function returns `Conflict` rather than scheduling the orchestration.

**Example test case:**

Arrange:

- folder expansion returns `Cases/123/Source/a.pdf`
- computed destination is `Cases/123/Dest/Source/a.pdf`
- destination existence check returns `true`

Assert:

- function returns `Conflict`
- `ScheduleNewOrchestrationInstanceAsync` is not called

backend/CPS.ComplexCases.FileTransfer.API.Tests/Unit/Functions/CleanupStaleManageMaterialsOperationsTests.cs
```diff
@@ -0,0 +1,377 @@
...
+            _loggerMock.Object,
+            _serviceMock.Object,
+            Options.Create(_config));
+    }
+
+    // ── Age-based deletion ────────────────────────────────────────────
+
+    [Fact]
+    public async Task Run_WhenRowExceedsMaxAge_DeletesWithoutCheckingDurableStatus()
+    {
```
HasanCPS, line 48: The cleanup tests currently assert old rows are deleted by age. Once the cleanup behavior is fixed, we should assert that a stale-aged row is retained when the durable orchestration is still `Running`.

**Example test case:**

Arrange:

- active operation `CreatedAt` is older than `MaxAgeHours`
- durable instance status is `Running`

Assert:

- `DeleteOperationAsync` is not called

backend/CPS.ComplexCases.FileTransfer.API/Durable/Activity/WriteCopyActivityLog.cs
HasanCPS: **Please add tests for folder activity-log outcomes where all files succeed, no files succeed, and only some files succeed.**

This protects against partial folder copies being logged as fully `Copied`.

**Example test case:**

Arrange:

- original operation is `Folder` with source path `Cases/123/FolderA/`
- expanded folder had `a.pdf` and `b.pdf`
- `SuccessfulSourceKeys` contains only `Cases/123/FolderA/a.pdf`

Assert:

- activity log details do not mark the folder as fully `Copied`

backend/CPS.ComplexCases.FileTransfer.API/Functions/InitiateBatchCopy.cs
```diff
@@ -0,0 +1,314 @@
...
+            Id = transferId,
+            CaseId = request.CaseId,
+            OperationType = "BatchCopy",
+            SourcePaths = JsonSerializer.Serialize(sourcePaths),
+            DestinationPaths = JsonSerializer.Serialize(destinationPaths),
+            UserName = request.UserName,
+            CreatedAt = now,
+        };
+
+        var inserted = await _caseActiveManageMaterialsService.CheckConflictAndInsertAsync(
```
HasanCPS, line 277: The new `CheckConflictAndInsertAsync` flow fixes the check-then-insert race, but the row is still inserted before `ScheduleNewOrchestrationInstanceAsync`. If scheduling throws after the insert succeeds, the API returns an error but the active-operation row remains with no orchestration behind it. With the updated cleanup behavior, a missing durable instance is retained until `MaxAgeHours`, so this can block manage-materials operations on the case for that whole window.

**Example:**

- request passes validation and NetApp preflight
- `CheckConflictAndInsertAsync` inserts a `case_active_manage_materials` row
- `ScheduleNewOrchestrationInstanceAsync` then throws due to a transient Durable Functions/storage issue

Current behavior:

- the function fails before returning `Accepted`
- no copy orchestration exists
- the active-operation row remains
- later copy/move/delete requests can get `Conflict` until cleanup eventually removes the orphaned row

Expected behavior:

- if scheduling fails after the lock row is inserted, the inserted row is removed or otherwise reconciled immediately
- the case is not blocked by an orphaned active-operation row

**Suggestion:** wrap `ScheduleNewOrchestrationInstanceAsync` in a `try/catch` and call `DeleteOperationAsync(transferId)` when scheduling fails, then rethrow/return the error. Add a unit test where `ScheduleNewOrchestrationInstanceAsync` throws and verify `DeleteOperationAsync(transferId)` is called.

backend/CPS.ComplexCases.API/Functions/GetActiveManageMaterialsOperations.cs
```diff
@@ -0,0 +1,60 @@
...
+        FunctionContext functionContext,
+        int caseId)
+    {
+        var context = functionContext.GetRequestContext();
+        _initializationHandler.Initialize(context.Username, context.CorrelationId, caseId);
+
+        _logger.LogInformation("GetActiveManageMaterialsOperations request for CaseId: {CaseId}. CorrelationId: {CorrelationId}",
+            caseId, context.CorrelationId);
+
+        var operations = await _caseActiveManageMaterialsService.GetActiveOperationsForCaseAsync(caseId);
```
HasanCPS, line 46: The endpoint has auth attributes, but it directly queries `GetActiveOperationsForCaseAsync(caseId)`. `GetCase` performs the extra CMS/DDEI lookup before returning case data, but this polling endpoint does not. If a caller can guess another case ID, this can leak active operation metadata such as source paths, destination paths, usernames, and created timestamps.

**Example:**

- user is authenticated for case `123`
- user calls `GET /v1/cases/999/manage-materials/active`
- case `999` currently has a copy/delete/move operation in progress

Current behavior:

- the endpoint queries rows by `caseId = 999`
- active operation metadata can be returned without the same case-access check used by `GetCase`

Expected behavior:

- the endpoint verifies case access before querying or returning active operation rows
- unauthorized case IDs return the same 403/404 behavior as other case endpoints

**Suggestion:** mirror `GetCase` by resolving the case via CMS/DDEI, or a shared authorization helper, before calling `GetActiveOperationsForCaseAsync`.

backend/CPS.ComplexCases.FileTransfer.API/Durable/Orchestration/CopyOrchestrator.cs
```diff
@@ -0,0 +1,278 @@
...
+                    CaseId = input.CaseId,
+                    UserName = input.UserName,
+                    CorrelationId = input.CorrelationId,
+                    OriginalOperations = input.OriginalOperations,
+                    SuccessfulSourceKeys = successfulSourceKeys,
+                });
+
+            telemetryEvent.IsSuccessful = telemetryEvent.TotalFilesFailed == 0;
+            telemetryEvent.OrchestrationEndTime = DateTime.UtcNow;
+            _telemetryClient.TrackEvent(telemetryEvent);
```
HasanCPS: On the success path, `_telemetryClient.TrackEvent(telemetryEvent)` runs at the end of the `try`, then the `finally` block runs and tracks the same event again. That double-counts successful batch-copy orchestrations and can skew success counts, duration, and cost/volume metrics.

**Example:**

- all files copy successfully
- `telemetryEvent.IsSuccessful` is set to `true`
- the event is tracked once in the `try`
- `finally` updates `OrchestrationEndTime` and tracks the event again

Current behavior:

- one successful orchestration creates two telemetry events

Expected behavior:

- one orchestration creates one orchestration-level telemetry event

**Suggestion:** track the orchestration event in exactly one place, for example only in `finally` after setting `IsSuccessful` and `OrchestrationEndTime`, or only at the end of the successful path with no duplicate in `finally`.

backend/CPS.ComplexCases.FileTransfer.API/Validators/CopyNetAppBatchRequestValidator.cs
```diff
@@ -0,0 +1,32 @@
...
+
+public class CopyNetAppBatchRequestValidator : AbstractValidator<CopyNetAppBatchRequest>
+{
+    public CopyNetAppBatchRequestValidator()
+    {
+        RuleFor(x => x.CaseId)
+            .GreaterThan(0)
+            .WithMessage("CaseId must be greater than 0.");
+
+        RuleFor(x => x.DestinationPrefix)
```
HasanCPS, line 17: The public API validator blocks path traversal (`..`), leading `/`, duplicate source paths, overlapping folder/file paths, folder-copy-into-self, same-key material copies, and batches over `MaxOperations`. The FileTransfer validator only checks required fields and that `DestinationPrefix` ends with `/`. If this function is ever called directly with its function key, or routed around the public API, those stronger rules are bypassed at the service that actually starts the orchestration.

**Example:**

- public API rejects a request containing duplicate `sourcePath` values or `../` path traversal
- the same request shape sent directly to FileTransfer passes this validator
- FileTransfer proceeds into NetApp preflight and durable orchestration with inputs the public API considers invalid

Current behavior:

- validation rules drift between the gateway API and the worker API

Expected behavior:

- FileTransfer enforces the same safety invariants before starting durable work

**Suggestion:** share the validator rules between the two request types, or duplicate the public API path/overlap/max-size rules in the FileTransfer validator as defense in depth.

backend/CPS.ComplexCases.FileTransfer.API/Durable/Activity/WriteCopyActivityLog.cs
```diff
@@ -0,0 +1,93 @@
...
+                outcome = successCount == op.ExpectedSourceKeys.Count ? "Copied"
+                        : successCount > 0 ? "Partial"
+                        : "NotCopied";
+            }
+            else
+            {
+                outcome = successfulKeySet.Contains(op.SourcePath) ? "Copied" : "NotCopied";
+            }
+
+            return new
```
HasanCPS, line 51: The existing transfer activity logs use `FileTransferDetails`, which records both `SourcePath` and `DestinationPath` for Egress-to-NetApp / NetApp-to-Egress transfers. The new batch copy log only writes `sourcePath`, `outcome`, and `type` per item, so the final activity log does not tell users or auditors where the item was copied to.

**Example:**

- user copies `Cases/123/FolderA/report.pdf`
- destination prefix is `Cases/123/Disclosure/`
- copy succeeds

Current activity log details:

- `sourcePath`: `Cases/123/FolderA/report.pdf`
- `outcome`: `Copied`
- `type`: `Material`
- no destination path or destination prefix

Expected behavior:

- activity log details include both the source and the destination, for example `destinationPath` or `destinationPrefix`
- users can tell what was copied and where it was copied to, similar to normal transfer activity logs

**Suggestion:** carry the computed destination path/prefix through the batch copy payload and include it in each activity log item, e.g. `sourcePath`, `destinationPath`, `outcome`, and `type`. For folder operations, include the destination folder prefix or expanded per-file destinations depending on the audit requirement.

ui-spa/src/components/case-management/activity-log/ActivityTimeline.tsx
```diff
@@ -184,6 +186,29 @@ export const ActivityTimeline: React.FC<ActivityTimelineProps> = ({
     </div>
   );
 
+  const renderBatchCopyDetails = (details: BatchCopyDetails) => (
```
HasanCPS, line 189: The backend now writes `destinationPath` for each batch copy item, but the UI type and renderer still only use `sourcePath`, `outcome`, and `type`. Users can see what was copied, but not where it was copied to, unlike normal transfer activity logs that show both source and destination.

**Example:**

- source path: `Cases/123/FolderA/report.pdf`
- destination path: `Cases/123/Disclosure/report.pdf`
- outcome: `Copied`

Current UI:

- shows `Cases > 123 > FolderA > report.pdf`
- shows `Copied`
- does not show `Cases > 123 > Disclosure > report.pdf`

Expected UI:

- shows the source path
- shows the destination path
- shows the outcome

**Suggestion:** add `destinationPath` to `BatchCopyItem` in `ActivityLogResponse.ts`, then update `renderBatchCopyDetails` to display both source and destination for each copied item, similar to the transfer activity log source/destination layout.

HasanCPS (approved):
LGTM!

PR #417, opened 2026-04-28
Task/fct2 16748 cleanup automated testing uploaded files

lilachdavis:


devops-pipelines/scripts/test-automation-cleanup/Clear-SharedDriveFolder.ps1
```diff
@@ -0,0 +1,127 @@
...
+# ============================================================
+# DELETE FOLDER
+# ============================================================
+Write-Host "Attempting to delete folder $FolderPath..." -ForegroundColor Yellow
+
+$body = @{
+  caseId = $CaseId
+  operations = @(
+    @{
+      type = "Folder"
```
HasanCPS, line 73: The NetApp cleanup sends `sourcePath = "$FolderPath/"`, but `DeleteNetAppBatch` only allows paths under the case's stored `NetappFolderPath`.

The pipeline passes `-FolderPath "integration-tests"`, so the request body contains `integration-tests/`. Unless the case metadata path is also exactly `integration-tests`, the API rejects it as outside the case folder and the cleanup exits before recreating the folder.

**Example:**

- input: `-CaseId $(IntegrationTestsCaseId) -FolderPath "integration-tests"`
- current behavior: sends `sourcePath = "integration-tests/"`, then `DeleteNetAppBatch` rejects it when it does not start with the case NetApp folder prefix.
- expected behavior: the cleanup deletes only the intended test folder under the connected case folder and then recreates it.

**Suggestion:** Build the full NetApp source path from the connected case folder path, or use a cleanup route/script that deletes by an environment-owned bucket prefix rather than the case-scoped batch delete API.

devops-pipelines/scripts/test-automation-cleanup/Remove-EgressWorkspacesByPrefix.ps1
```diff
@@ -0,0 +1,131 @@
...
+
+# ============================================================
+# LIST WORKSPACES
+# ============================================================
+Write-Host "[2/3] Listing Workspaces..." -ForegroundColor Yellow
+
+$workspacesToRemove = @()
+
+try {
+  $response = Invoke-RestMethod -Method Get `
```
HasanCPS, line 54: The script deletes every workspace returned by `workspaces?name=$WorkspaceNamePrefix` without verifying each returned workspace name starts with the prefix.

Because this is destructive and the pipeline passes `-Force`, the script should not rely solely on Egress search semantics. If `name=` is fuzzy, contains-based, or changes behavior, unrelated workspaces could be deleted.

**Example:**

- input: `-WorkspaceNamePrefix "AUTOMATION-TESTING"`
- current behavior: deletes every workspace returned by the Egress API search.
- expected behavior: deletes only workspaces whose names strictly match the automation-test naming pattern.

**Suggestion:** Filter `$workspacesToRemove` locally with a strict prefix or regex check before printing or deleting. Consider adding an age threshold as another guard for generated resources.

devops-pipelines/scripts/test-automation-cleanup/Clear-SharedDriveFolder.ps1
```diff
@@ -0,0 +1,127 @@
...
+} | ConvertTo-Json
+
+try {
+  $response = Invoke-RestMethod -Method Post `
+    -Uri "$BaseUrl/api/v1/netapp/delete/batch" `
+    -Headers $authHeader `
+    -ContentType 'application/json' `
+    -Body $body
+  
+  if ($response.status -eq 'Completed' -and $response.succeeded -gt 0) {
```
HasanCPS, line 86: A missing folder is treated as a failure, so the nightly cleanup can fail on an already-clean environment.

`DeleteNetAppBatch` can return `Status = "NoOp"` with `NotFound > 0` when the folder is already absent. For cleanup, that should usually be acceptable and the script should still recreate the folder. Currently it writes an error and exits before recreation.

**Example:**

- input: cleanup runs after the folder was already deleted or never created.
- current behavior: delete response is not `Completed`, so the script exits with failure.
- expected behavior: missing folder is accepted and the recreate step still runs.

**Suggestion:** Treat `Completed` and `NoOp` with zero failures as acceptable delete outcomes, then run the recreate step.

devops-pipelines/scripts/test-automation-cleanup/Clear-SharedDriveFolder.ps1
```diff
@@ -0,0 +1,127 @@
...
+  [string]$AadPassword,
+
+  [Parameter(Mandatory = $true)]
+  [int]$CaseId,
+
+  [Parameter(Mandatory = $true)]
+  [string]$FolderPath,
+
+  [Parameter(Mandatory = $false)]
+  [string]$BaseUrl = "https://fa-lacc-api-dev.azurewebsites.net"
```
HasanCPS: can we remove this URL?

lilachdavis: Is this a security concern? 
We have "fa-lacc-api-${{parameters.environment}}" elsewhere in the repo - does that also need obscuring?

devops-pipelines/scripts/test-automation-cleanup/Clear-SharedDriveFolder.ps1
lilachdavis: I don't follow why we would want to create a folder that did not exist in the first place. The purpose of this is to empty the contents of an existing folder. If the input for $FolderPath is the name of a non-existent folder, I would like to be alerted that I may have given the wrong input, rather than create a folder with the wrong name that shouldn't have existed in the first place.

What I consider 'an already clean environment' is that the folder itself exists, but is empty.
This is based on the way the integration tests are set up, where the netapp folder 'integration-tests' is expected to already exist within flexgroup4 and be associated with the specified CMS caseId. If a folder of that name doesn't exist for that case, the integration tests would also fail.

Based on that, I spoke with Kamesh, and we agreed that his E2E tests for LCC will follow the same pattern - use an existing case, that is already associated with a specific NetApp folder within flexgroup4 (named existingCaseAutomation).

devops-pipelines/scripts/test-automation-cleanup/Remove-EgressWorkspacesByPrefix.ps1
lilachdavis: For age threshold, do you mean files newer than 24 hours? My thinking being, any generated files will have been created in the 24 hours since the last cleanup

devops-pipelines/scripts/test-automation-cleanup/Clear-SharedDriveFolder.ps1
lilachdavis: I'm not sure I understand this comment / suggestion.
"expected behavior: the cleanup deletes only the intended test folder under the connected case folder and then recreates it." - This is how it seems to behave already?
I was using the delete batch endpoint at your recommendation, My understanding was that we deliberately don't expose an API point that directly removes a folder not associated with a caseId. 
Both the integration and E2E tests are using folders within flexgroup4 that are connected to a known caseId, so I don't quite see the potential for failure here?

devops-pipelines/scripts/test-automation-cleanup/Remove-EgressWorkspacesByPrefix.ps1
lilachdavis: Added validation against the prefix.

HasanCPS (approved):
LGTM!

PR #422, opened 2026-05-05
Story/FCT2-17749 -  Zod schema validator

renjithabby:
https://cpsgovuk.atlassian.net/browse/FCT2-17749


ui-spa/src/apis/gateway-api.ts
```diff
@@ -299,35 +349,40 @@ export const getNetAppFolders = async (
...
+  return getNetAppFolders(
+    folderPath,
+    take,
+    pagination.nextContinuationToken,
+    updatedFolders,
+    updatedFiles,
+  );
 };
 
 export const indexingFileTransfer = async (
```
HasanCPS, line 397: **Invalid Transfer Payloads Are Still Sent**

Both `indexingFileTransfer` and `initiateFileTransfer`  (Line `377-429`) call `safeParse`, but the failure branch only writes a warning. The function then continues and posts the original `payload`.

**Example:**

- input: a payload missing `caseId`
- current behavior: `console.warn(...)` runs, then `fetch(...)` still sends the invalid request
- expected behavior: the client stops before the network call and surfaces a validation error

The current tests lock in the weaker behavior: the invalid request cases only assert the warning, not that the call is rejected or that `fetch` is skipped.

**Suggestion:** Return or throw on request validation failure before building the request. The tests should assert that invalid payloads reject and that `fetch` is not called.

ui-spa/src/apis/gateway-api.ts
```diff
@@ -100,20 +143,20 @@ export const getEgressSearchResults = async (
...
+
+  const { data, pagination } = result;
+  const updated = collected.concat(data);
+  if (skip + take >= pagination.totalResults) {
+    return updated;
   }
+  return getEgressSearchResults(workspaceName, skip + take, take, updated);
 };
 
 export const connectEgressWorkspace = async ({
```
HasanCPS, line 166: **The new connection request schemas are defined and exported, but the connection APIs never use them**

`connectEgressPayloadSchema` and `connectNetAppPayloadSchema` ( lines `162-264`) look intended to validate the POST bodies for `connectEgressWorkspace` and `connectNetAppFolder`. Those methods still build the JSON inline and send it directly. If `caseId` cannot be parsed, `JSON.stringify` turns `NaN` into `null`, so the client can still send a shape that the schema would reject.

**Example:**

- input: `caseId: "abc"`
- current behavior: `parseInt("abc")` produces `NaN`, and the request body serializes it as `"caseId": null`
- expected behavior: the client catches the invalid payload before posting

**Suggestion:** Validate the constructed connection payloads with these schemas, or remove the unused schemas if connection validation is out of scope for this PR.

ui-spa/src/apis/gateway-api.ts
renjithabby: resolved

ui-spa/src/apis/gateway-api.ts
renjithabby: resolved

HasanCPS (approved):
LGTM!

PR #423, opened 2026-05-06
Feature/fct2 16592 move

bridgesr:


backend/CPS.ComplexCases.FileTransfer.API/Durable/Orchestration/MoveOrchestrator.cs
```diff
@@ -0,0 +1,322 @@
...
+
+                var retryResults = await Task.WhenAll(retryBatch);
+                await ProcessTransferResults(context, entityId, retryResults, telemetryEvent, isRetry: true);
+
+                allResults.RemoveAll(r => r != null && !r.IsSuccess && r.FailedItem?.ErrorCode == TransferErrorCode.Transient);
+                allResults.AddRange(retryResults);
+            }
+
+            // 5. Delete phase — remove source keys for all successfully copied files
+            await context.CallActivityAsync(
```
HasanCPS, line 96: **Issue:** source delete failures are still reported as successful moves.

The move flow treats copy success as the only input for the final outcome. `DeleteNetAppFiles` records failed source deletes, but it does not fail the orchestration. `MoveOrchestrator` then finalizes the transfer and passes only successful copy source keys into `WriteMoveActivityLog`.

**Example:**

- input: move `CaseRoot/A/file.pdf` to `CaseRoot/B/file.pdf`
- current behavior: if the copy succeeds but the source delete fails because the file is locked, the activity log can still show `Moved`
- expected behavior: the move should be marked partial/failed, or the activity log should show that the source still exists

**Suggestion:** feed delete failures into the final transfer status/activity log, or fail/partially complete the move when any copied source object cannot be deleted. Please add a test where `DeleteNetAppFiles` returns a deletion error and verify that the final status/activity details do not claim a clean move.

backend/CPS.ComplexCases.FileTransfer.API/Functions/InitiateBatchMove.cs
```diff
@@ -0,0 +1,320 @@
...
+                var expandedFiles = new List<(string SourceKey, string RelativeKey)>();
+                string? continuationToken = null;
+                do
+                {
+                    var listArg = _netAppArgFactory.CreateListObjectsInBucketArg(
+                        request.BearerToken, request.BucketName,
+                        continuationToken: continuationToken,
+                        prefix: sourcePrefix);
+
+                    var listResult = await _netAppClient.ListObjectsInBucketAsync(listArg);
```
HasanCPS: **Issue:** a partial folder listing can lead to deleting files that were never copied.

During folder expansion, `ListObjectsInBucketAsync` returning `null` breaks the pagination loop instead of failing the request. If this happens after at least one page has been expanded, the orchestration is scheduled with only that partial set of files. Later, `MoveOrchestrator` decides the folder can be cleaned up when all scheduled keys succeeded, and `DeleteNetAppSourceFolders` deletes the whole source folder prefix.

**Example:**

- input: move folder `CaseRoot/A/` containing files on two list pages
- current behavior: page 1 is copied, page 2 returns `null`, the move is scheduled for page 1 only, then folder cleanup can delete the full `CaseRoot/A/` prefix
- expected behavior: if any source expansion page cannot be listed, the move should not be scheduled and no source cleanup should run

**Suggestion:** treat a `null` source listing page as a preflight failure instead of `break`, and add a test where folder expansion succeeds for page 1 but returns `null` for page 2. The expected result should be an error response and no orchestration.

backend/CPS.ComplexCases.FileTransfer.API/Functions/InitiateBatchMove.cs
```diff
@@ -0,0 +1,320 @@
...
+
+                var destArg = _netAppArgFactory.CreateGetObjectArg(request.BearerToken, request.BucketName, computedDest);
+                var destExists = await _netAppClient.DoesObjectExistAsync(destArg);
+                if (destExists)
+                {
+                    preflight409Errors.Add($"A file already exists at the destination: {computedDest}");
+                    continue;
+                }
+
+                var caseCheckArg = _netAppArgFactory.CreateListObjectsInBucketArg(
```
HasanCPS: **Issue:** material destination clash detection only checks the first list page.

The folder destination check paginates through the destination prefix, but the material case-insensitive clash check calls `ListObjectsInBucketAsync` once. If the destination folder contains more objects than one NetApp/S3 page, a conflicting key on a later page can be missed.

**Example:**

- input: move `CaseRoot/Source/report.pdf` to `CaseRoot/Dest/`
- current behavior: if `CaseRoot/Dest/REPORT.PDF` is returned on page 2, the move can proceed
- expected behavior: the move should return a conflict before scheduling the orchestration

**Suggestion:** use the same continuation-token loop used by the folder destination scan, then check every returned destination key for a case-insensitive filename clash. Add a unit test where the clash appears after the first page.

backend/CPS.ComplexCases.FileTransfer.API/Functions/InitiateBatchMove.cs
```diff
@@ -0,0 +1,353 @@
...
+
+                var existingDestKeys = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+                string? destContinuationToken = null;
+                do
+                {
+                    var destListArg = _netAppArgFactory.CreateListObjectsInBucketArg(
+                        request.BearerToken, request.BucketName,
+                        continuationToken: destContinuationToken,
+                        prefix: destFolderPrefix);
+                    var destListResult = await _netAppClient.ListObjectsInBucketAsync(destListArg);
```
HasanCPS: **Issue:** destination listing failures still fail open, so the preflight check can miss existing destination objects.

The source folder listing now fails the request when a page returns `null`, which is good. The destination checks do not do the same. In the folder destination loop, `destListResult == null` breaks the loop and schedules the move with an incomplete `existingDestKeys` set. In the material case-insensitive clash loop, `caseCheckResult == null` also breaks and treats the partial scan as clean.

**Example:**

- input: move `CaseRoot/A/` to `CaseRoot/B/`, where `CaseRoot/B/A/file.pdf` exists on a later destination listing page
- current behavior: if that destination listing page returns `null`, the move can still be scheduled
- expected behavior: the move should fail preflight rather than risk overwriting or colliding with an existing NetApp object

**Suggestion:** treat `null` destination listing pages as preflight listing errors, the same way source listing errors are handled. Add tests for a `null` destination folder page and a `null` material clash page.

backend/CPS.ComplexCases.FileTransfer.API/Functions/InitiateBatchMove.cs
```diff
@@ -0,0 +1,430 @@
...
+        var preflight404Errors = new List<string>();
+        var preflightListingErrors = new List<string>();
+
+        foreach (var op in request.Operations)
+        {
+            var result = string.Equals(op.Type, "Folder", StringComparison.OrdinalIgnoreCase)
+                ? await ProcessFolderOperationAsync(op, request, correlationId)
+                : await ProcessFileOperationAsync(op, request, correlationId);
+
+            moveFileItems.AddRange(result.FileItems);
```
HasanCPS, line 106: **Reject duplicate destinations within a move batch**

  The batch preflight should reject requests where two operations resolve to the same destination key before the durable workflow is scheduled.

  ### Example

  
  ```
A/report.pdf -> Dest/report.pdf
  B/report.pdf -> Dest/report.pdf
```

  Today, each operation checks the destination independently. Both operations see that Dest/report.pdf does not exist yet, so both are allowed. During execution, one copy can overwrite the other, and then both source files are deleted.

  That creates a data-loss path.

  ### Suggested fix

  After resolving `DestinationPrefix` and `DestinationFileName` for every operation, build the final destination key for each move and reject the request if any duplicate keys exist in the same batch.

  The duplicate check should use the same normalization rules as the actual move, including slash handling and case sensitivity.

File: backend/CPS.ComplexCases.FileTransfer.API/Durable/Activity/DeleteNetAppSourceFolders.cs [hunk 6]
```diff
@@ -0,0 +1,61 @@
...
+        foreach (var folderPath in payload.SourceFolderPaths)
+        {
+            try
+            {
+                var arg = _netAppArgFactory.CreateDeleteFileOrFolderArg(
+                    payload.BearerToken,
+                    payload.BucketName,
+                    "DeleteSourceFolder",
+                    folderPath,
+                    isFolder: true);
```
HasanCPS: **Avoid recursively deleting moved source folders**

  Folder cleanup should not recursively delete the original source folder unless it has confirmed that nothing new appeared there after the move was expanded.

  ### Example

  1. The move expands Folder/ and finds:
     Folder/a.txt

  2. Folder/a.txt is copied to the destination.

  3. Before cleanup runs, another process writes:
     Folder/new-file.txt

  4. Cleanup recursively deletes Folder/

  Folder/new-file.txt was never copied, but it can still be deleted by the recursive cleanup. This is especially risky because NetApp folders may also be modified outside this API, for example through SMB.

  ### Suggested fix

  Only delete the specific source file keys that were moved.

  For folder roots, either:

  - delete only a known empty folder marker, or
  - re-list the source prefix immediately before cleanup and skip deleting the folder if any unexpected keys remain.

backend/CPS.ComplexCases.FileTransfer.API/Functions/InitiateBatchMove.cs
```diff
@@ -0,0 +1,430 @@
...
+                bearerToken, bucketName,
+                continuationToken: continuationToken,
+                prefix: destinationPrefix);
+            var result = await _netAppClient.ListObjectsInBucketAsync(arg);
+
+            if (result == null) return (false, true);
+
+            if (result.Data.FileData.Any(f =>
+                    string.Equals(Path.GetFileName(f.Path), fileName, StringComparison.OrdinalIgnoreCase)
+                    && !string.Equals(f.Path, computedDest, StringComparison.Ordinal)))
```
HasanCPS, line 419: **Ignore nested files when checking move destination clashes**

  The destination clash check should not reject a valid move just because a matching filename exists inside a nested destination folder.

  ### Example

  Existing file:
  Dest/sub/REPORT.PDF

  Requested move:
  Source/report.pdf -> Dest/report.pdf

  These are different destination keys, so the move should be allowed. However, the current check lists every object under Dest/ and compares only Path.GetFileName(...). Both files appear to have the same filename, so the request is rejected incorrectly.

  ### Suggested fix

  Compare the full computed destination key against existing keys, using the same case-sensitivity rules as the storage layer.

  Alternatively, only check files directly inside the immediate destination folder and ignore descendants under nested folders.

backend/CPS.ComplexCases.FileTransfer.API/Durable/Activity/DeleteNetAppSourceFolders.cs
```diff
@@ -0,0 +1,116 @@
...
+
+                if (hasUnexpected)
+                {
+                    _logger.LogWarning(
+                        "Source folder {FolderPath} contains unexpected files after move. Skipping recursive deletion to avoid data loss. TransferId: {TransferId}.",
+                        folderSpec.FolderPath, payload.TransferId);
+                    continue;
+                }
+
+                var arg = _netAppArgFactory.CreateDeleteFileOrFolderArg(
```
HasanCPS, line 60: Folder cleanup still uses recursive delete after the safety check

  `DeleteNetAppSourceFolders` now lists the source folder before cleanup, but it still calls the recursive folder delete path:

` CreateDeleteFileOrFolderArg(..., folderSpec.FolderPath, isFolder: true)
`

That still leaves a data-loss window. `DeleteFileOrFolderAsync(... isFolder: true)` re-lists everything under the prefix and deletes all matching keys. So if another process writes a file after `HasUnexpectedKeysAsync` runs but before the recursive delete lists the folder, that new file can still be deleted even though it was never copied.

  ### Example

  1. Folder/old.txt is copied and deleted successfully.
  2. DeleteNetAppSourceFolders lists Folder/ and sees it is safe.
  3. SMB writes Folder/new.txt.
  4. Recursive folder delete runs and deletes Folder/new.txt.

There is also a same-key case: if Folder/old.txt is recreated after the individual source delete, the current `ExpectedSourceKeys` check treats that key as expected and allows the recursive delete.

  ### Suggested fix

  Do not call the recursive folder delete path for move cleanup.

  After the moved source files have already been deleted individually, cleanup should either:

  - delete only a known folder marker key non-recursively, or
  - skip folder cleanup whenever the final source-prefix listing returns any file object at all.

  If the NetApp client only exposes recursive folder deletion today, this probably needs a separate non-recursive folder-marker delete path rather than reusing isFolder: true.

backend/CPS.ComplexCases.FileTransfer.API/Functions/InitiateBatchMove.cs
```diff
@@ -0,0 +1,443 @@
...
+        {
+            var listArg = _netAppArgFactory.CreateListObjectsInBucketArg(
+                bearerToken, bucketName,
+                continuationToken: continuationToken,
+                prefix: sourcePrefix);
+            var listResult = await _netAppClient.ListObjectsInBucketAsync(listArg);
+
+            if (listResult == null) return (files, true);
+
+            foreach (var file in listResult.Data.FileData)
```
HasanCPS, line 357: **Folder marker objects can be moved to the wrong destination**

`ExpandSourceFilesAsync` adds every listed `FileData` item, including the folder marker key itself. NetApp list results map all S3 objects into `FileData `(`backend/CPS.ComplexCases.NetApp/Client/NetAppClient.cs:250`), so a marker like `Case/OldFolder/` is included with `relativeKey == ""`.

Example:

 ```
 Source folder: Case/OldFolder/
 DestinationPrefix: Case/New/
  Listed objects:
  - Case/OldFolder/
  - Case/OldFolder/report.pdf
```

  The marker becomes a move item with:

  ```
DestinationPrefix = Case/New/OldFolder/
DestinationFileName = ""
```

But `TransferFile` treats an empty ModifiedPath as “not set” (
`backend/CPS.ComplexCases.FileTransfer.API/Durable/Activity/ TransferFile.cs:59`), so the marker is uploaded using the original source key and lands under something like:

 ` Case/New/OldFolder/Case/OldFolder/`

For an empty-folder move, this means the destination folder is created in the wrong place. For a non-empty folder, it creates an extra nested marker. The move expansion should either skip the source folder marker and handle destination marker creation explicitly, or make empty `DestinationFileName` a valid relative destination.

backend/CPS.ComplexCases.FileTransfer.API/Durable/Activity/DeleteNetAppSourceFolders.cs
```diff
@@ -0,0 +1,117 @@
...
+        {
+            var arg = _netAppArgFactory.CreateListObjectsInBucketArg(
+                bearerToken, bucketName,
+                continuationToken: continuationToken,
+                prefix: folderPrefix);
+            var result = await _netAppClient.ListObjectsInBucketAsync(arg);
+
+            if (result == null) return (false, true);
+
+            if (result.Data.FileData.Any())
```
HasanCPS: **Empty-folder cleanup treats the folder marker as a remaining file**

The new cleanup lists the folder prefix and skips deletion if `FileData.Any()`. For an empty folder with only the marker object left, NetApp can return that marker as `FileData`, for example:

  `Case/OldFolder/`

That makes `hasFiles == true`, so the activity skips deleting the exact marker it was created to remove. The check should ignore the marker key equal to `folderMarkerKey` and only block deletion when there are other objects below the prefix.

backend/CPS.ComplexCases.FileTransfer.API/Functions/InitiateBatchMove.cs
```diff
@@ -0,0 +1,452 @@
...
+
+        if (destPaginationFailed)
+        {
+            _logger.LogError(
+                "Destination folder listing returned null mid-pagination for {DestFolderPrefix}. CaseId: {CaseId}. CorrelationId: {CorrelationId}",
+                destFolderPrefix, request.CaseId, correlationId);
+            return EmptyResult(listingError: $"Failed to list destination folder contents: {destFolderPrefix}");
+        }
+
+        var (fileItems, errors409, scheduledFolderKeys) = BuildFolderMoveItems(expandedFiles, destFolderPrefix, existingDestKeys);
```
HasanCPS, line 281: **Empty folder move is silently ignored when batched with valid items**

The latest marker fix skips the source folder marker during expansion, which fixes the bad marker transfer for non-empty folders.
 But if a selected folder contains only its marker, `ProcessFolderOperationAsync` now returns no `FileItems` and no error.

  The whole-batch guard only rejects when **every** operation expands to zero files, so this can be silently ignored when the request also contains a valid move.

Example:

 ```
 Operations:
  - Folder: Case/EmptyFolder/   // only marker exists
  - Material: Case/report.pdf   // valid
```

 The API accepts the batch and schedules only `Case/report.pdf`; `Case/EmptyFolder/` is left behind even though the user selected it for move.

  The folder operation should either return a per-operation error after marker filtering leaves `expandedFiles.Count == 0`, or explicitly support empty-folder moves by creating/deleting the folder marker.

HasanCPS (approved):
Looks good to me, pending the backend integration tests once the MDS issues are sorted.

PR #425, opened 2026-05-06
fix issue exceeding command-line length limit

lilachdavis:
The Issue: 
When passing an array of > 100 IDs, Clear-EgressFolder script was failing to execute the Remove-EgressFiles function due to exceeding the command-line length limit.
The Fix:
Execute the function for batches of 100 IDs or under.


HasanCPS (approved):
LGTM!

PR #426, opened 2026-05-06
Story/FCT2- 17422 Tree View component

renjithabby:
https://cpsgovuk.atlassian.net/browse/FCT2-17422


ui-spa/src/components/common/tree-view-component/TreeView.tsx
```diff
@@ -0,0 +1,345 @@
...
+        ref={(el) => {
+          refs.current[node.id] = el;
+        }}
+        aria-expanded={node.isFolder ? isExpanded : undefined}
+        aria-busy={isLoading ? true : undefined}
+        aria-level={level}
+        id={node.id}
+        data-testid={node.id}
+        aria-selected={isSelected}
+        aria-labelledby={`name-${node.id}`}
```
HasanCPS: **Issue:** The component builds `aria-labelledby` and DOM IDs directly from `node.id`, which is a file/folder path. Paths can contain spaces, and `aria-labelledby` treats spaces as separators between multiple IDs.

**Location:** `ui-spa/src/components/common/tree-view-component/TreeView.tsx`, lines `268` and `293`

`aria-labelledby={`name-${node.id}`}` and `id={`name-${node.id}`}` assume `node.id` is safe to use as an ID reference. For a path like `folder a/sub folder`, the generated label ID contains spaces. In `aria-labelledby`, that value is split into multiple ID references, so assistive tech can fail to resolve the intended label.

**Example:**

- input: node id/path = `Cases/Folder A/Sub Folder`
- current behavior: generated label ID contains spaces; `aria-labelledby` is split into multiple unresolved ID references
- expected behavior: ARIA references should use a sanitized, stable ID that does not include whitespace or unsafe characters

**Suggestion:** derive a safe DOM ID (for example, by encoding or hashing `node.id`) and use that value consistently for `id`, `aria-labelledby`, and related attributes.

ui-spa/src/components/common/tree-view-component/TreeViewComponent.tsx
```diff
@@ -0,0 +1,345 @@
...
+        }
+      }
+      return undefined;
+    },
+    [loadedChildren],
+  );
+
+  const handleLoadingNodeChildren = useCallback(
+    (node: TreeNode, id: string) => {
+      const hasChildren =
```
HasanCPS, line 99: **Issue:** The lazy-load guard treats an empty loaded folder as "not loaded", so `onLoadChildren` is called again each time the folder is expanded.

The current check uses `loadedChildren[id].length > 0`. If the API returns `[]` (valid empty folder), this stays false forever, so repeated toggles keep reloading the same folder.

**Example:**

- input: expand folder `X`, API returns `[]`, collapse, expand again
- current behavior: `onLoadChildren("X")` runs again
- expected behavior: folder should be considered loaded after first response, even when empty

**Suggestion:** track "loaded" by key presence (`id in loadedChildren`) or with an explicit `loadedIds` set, instead of checking `length > 0`.

renjithabby: fixed

HasanCPS (approved):
lgtm!

PR #430, opened 2026-05-11
change cleanup pipeline to only trigger on weekdays

lilachdavis:
Setting cleanup cron job to only trigger early AM Tue-Sat, as no tests are expected to run over the weekend


HasanCPS (approved):
LGTM!

PR #433, opened 2026-05-11
Story/FCT2-16331 - Disconnect shared drive folder

renjithabby:
https://cpsgovuk.atlassian.net/browse/FCT2-16331

### DIFF
--- a/ui-spa/src/components/case-management/netapp-disconnect/DisconnectSharedDriveConfirmationPage.tsx
+++ b/ui-spa/src/components/case-management/netapp-disconnect/DisconnectSharedDriveConfirmationPage.tsx
@@ -25,6 +25,8 @@ const DisconnectSharedDriveConfirmationPage = () => {
   };
   const errorSummaryRef = useRef<HTMLInputElement>(null);
 
+  const [disableButtons, setDisableButtons] = useState(false);
+
   const [formData, setFormData] = useState<{
     disconnectSharedDriveRadio?: GeneralRadioValue;
   }>({
@@ -37,7 +39,7 @@ const DisconnectSharedDriveConfirmationPage = () => {
     (errorKey: keyof FormDataErrors) => {
       return {
         children: formDataErrors[errorKey]?.errorSummaryText,
-        href: "#disconnect-shared-drive-yes",
+        href: "#disconnect-shared-drive-radio-yes",
         "data-testid": "disconnect-shared-drive-radio-link",
       };
     },
@@ -96,8 +98,9 @@ const DisconnectSharedDriveConfirmationPage = () => {
       navigate(`/case/${caseId}/case-management`);
       return;
     }
-
+    setDisableButtons(true);
     const response = await disconnectNetAppFolder(caseId);
+    setDisableButtons(false);
     if (!response.ok) {
       navigate(
         `/case/${caseId}/case-management/disconnect-shared-drive-failure`,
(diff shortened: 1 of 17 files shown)


ui-spa/src/components/case-management/netapp-disconnect/DisconnectSharedDriveConfirmationPage.tsx
```diff
@@ -0,0 +1,184 @@
...
+import { Radios, Button, ErrorSummary } from "../../govuk";
+import { useNavigate, useLocation, Link } from "react-router-dom";
+import { disconnectNetAppFolder } from "../../../apis/gateway-api";
+import styles from "./DisconnectSharedDriveConfirmationPage.module.scss";
+
+type GeneralRadioValue = "yes" | "no" | "";
+
+const DisconnectSharedDriveConfirmationPage = () => {
+  const navigate = useNavigate();
+  const {
```
HasanCPS, line 11: The new disconnect route destructures `useLocation().state` directly. If the page is opened directly, restored from browser history, or bookmarked, `location.state` can be `null`, which throws before the page renders.

Example route entry:

  ```text
  /case-management/123/netapp-disconnect
  ```

Since the URL already contains `:caseId`, read `caseId` from `useParams()` and treat `location.state` as optional. The success and failure pages should use the same guard pattern or redirect like the other guarded routes.

ui-spa/src/components/case-management/netapp-disconnect/DisconnectSharedDriveConfirmationPage.tsx
```diff
@@ -0,0 +1,184 @@
...
+  const handleSubmit = async (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (!validateFormData()) return;
+
+    if (formData.disconnectSharedDriveRadio === "no") {
+      navigate(`/case/${caseId}/case-management`);
+      return;
+    }
+    setDisableButtons(true);
```
HasanCPS, line 112: The submit handler handles non-OK responses, but it does not handle thrown errors from token acquisition or `fetch`. If the gateway is unavailable, CORS fails, or the network request rejects, the handler exits before re-enabling buttons or navigating to the failure page.

Wrap the request in `try/catch/finally` so thrown errors follow the same failure path as non-OK responses and the disabled state is always cleared.

ui-spa/src/components/case-management/netapp-disconnect/DisconnectSharedDriveConfirmationPage.tsx
renjithabby: fixed

ui-spa/src/components/case-management/netapp-disconnect/DisconnectSharedDriveConfirmationPage.tsx
renjithabby: fixed

ui-spa/src/components/case-management/netapp-disconnect/DisconnectSharedDriveConfirmationPage.tsx
```diff
@@ -0,0 +1,204 @@
...
+    try {
+      const response = await disconnectNetAppFolder(caseId);
+      if (!response.ok) {
+        navigate(
+          `/case/${caseId}/case-management/disconnect-shared-drive-failure`,
+          { state: { caseId, urn, isRouteValid: true } },
+        );
+        return;
+      }
+    } catch (e) {
```
HasanCPS, line 122: **Rejected disconnect requests still navigate to the success page**

The new `catch` handles rejected disconnect requests by navigating to the failure page, but it does not return afterward. After the `finally` block runs, execution continues to the success navigation at the bottom of `handleSubmit`.

That means failures from token acquisition, network errors, CORS errors, or a rejected `fetch` can briefly go through the failure path and then end on:

  ```text
  /case/{caseId}/case-management/disconnect-shared-drive-success
  ```

So the UI can tell the user the Shared Drive was disconnected even though the disconnect request never completed.

  Add a `return` after the failure navigation in the `catch`, matching the existing non-OK response branch.

  ```ts
  } catch (e) {
    console.error(e);
    navigate(
      `/case/${caseId}/case-management/disconnect-shared-drive-failure`,
      { state: { caseId, urn, isRouteValid: true } },
    );
    return;
  } finally {
    setDisableButtons(false);
  }
  ```

renjithabby: fixed

HasanCPS (approved):
LGTM!

PR #434, opened 2026-05-12
Feature/fct2 17888 preview material

bridgesr:


backend/CPS.ComplexCases.API/Services/DocumentService.cs
```diff
@@ -0,0 +1,128 @@
...
+{
+    private readonly ILogger<DocumentService> _logger = logger;
+    private readonly INetAppClient _netAppClient = netAppClient;
+    private readonly INetAppArgFactory _netAppArgFactory = netAppArgFactory;
+    private readonly IConversionService _conversionService = conversionService;
+    private readonly BlobServiceClient _blobServiceClient = blobServiceClient;
+
+    public async Task<FileStreamResult?> GetMaterialPreviewAsync(string path, string bearerToken, string bucketName)
+    {
+        var fileName = Path.GetFileName(path);
```
HasanCPS, line 26: **Use per-request temp blob names to avoid preview cross-talk**

The preview flow uses only `Path.GetFileName(path)` as the temp blob key, so different NetApp paths with the same filename share the same blobs:

  ```text
  tmp_statement.pdf
  preview_statement.pdf.pdf

  Example:

  /case-a/materials/statement.pdf
  /case-b/materials/statement.pdf

 ```
If two previews run at the same time, one request can overwrite or delete the other request’s `temp/preview` blob. That can return the wrong PDF preview or fail intermittently. Use a unique per-request prefix, for example a `GUID/correlation ID` plus the file name, and use that same key for save, convert, download, and cleanup.

backend/CPS.ComplexCases.API/Services/ConversionService.cs
```diff
@@ -0,0 +1,418 @@
...
+        var pdfBlobName = $"preview_{tmpFileName}.pdf";
+        var containerClient = await GetBlobContainerClientAsync();
+        var sourceBlob = containerClient.GetBlobClient($"tmp_{tmpFileName}");
+
+        using var pptStream = new MemoryStream();
+        await sourceBlob.DownloadToAsync(pptStream);
+        pptStream.Position = 0;
+
+        using var presentation = new Aspose.Slides.Presentation(pptStream);
+        using var slidePresentation = new Aspose.Slides.Presentation();
```
HasanCPS, line 232: **PowerPoint previews include the default blank slide**

`new Aspose.Slides.Presentation()` creates a presentation with a default blank slide. The code then clones the requested slide(s) into it without removing that default slide, so PPT/PPTX previews can start with a blank page before the actual material.

 Remove the default slide before adding clones:

  ```csharp
  using var slidePresentation = new Aspose.Slides.Presentation();
  slidePresentation.Slides.RemoveAt(0);

HasanCPS (approved):
LGTM!
The PPTX file conversion isn’t working as expected. It looks like a known issue, so We’ll handle it separately from this PR.

PR #435, opened 2026-05-14
Task/FCT2-18270- added protectedRoutes 

renjithabby:
https://cpsgovuk.atlassian.net/browse/FCT2-18270

### DIFF
--- a/ui-spa/src/common/utils/getUrlSearchParam.ts
+++ b/ui-spa/src/common/utils/getUrlSearchParam.ts
@@ -1,0 +1,6 @@
+export const getUrlSearchParam = (
+  property: string,
+  value: string | null | undefined,
+) => {
+  return new URLSearchParams({ [property]: value ?? "`" });
+};
--- a/ui-spa/src/common/utils/getUrlSearchParam.ts
+++ b/ui-spa/src/common/utils/getUrlSearchParam.ts
@@ -2,5 +2,5 @@ export const getUrlSearchParam = (
   property: string,
   value: string | null | undefined,
 ) => {
-  return new URLSearchParams({ [property]: value ?? "`" });
+  return new URLSearchParams({ [property]: value ?? "" });
 };
--- a/ui-spa/src/components/egress-connect/EgressConnectConfirmationPage.tsx
+++ b/ui-spa/src/components/egress-connect/EgressConnectConfirmationPage.tsx
@@ -40,18 +40,18 @@ const EgressConnectConfirmationPage: React.FC = () => {
           workspaceId: state?.selectedWorkspace.id,
           caseId: state?.caseId,
         });
-      }
 
-      if (!state?.isNetAppConnected)
-        navigate(
-          `/case/${state?.caseId}/netapp-connect?operation-name=${state?.selectedWorkspace.name}`,
-          {
-            state: {
-              searchQueryString: "",
+        if (!state?.isNetAppConnected)
+          navigate(
+            `/case/${state?.caseId}/netapp-connect?operation-name=${state?.selectedWorkspace.name}`,
+            {
+              state: {
+                searchQueryString: "",
+              },
             },
-          },
-        );
-      else navigate(`/case/${caseId}/case-management`);
+          );
+        else navigate(`/case/${caseId}/case-management`);
+      }
       // eslint-disable-next-line @typescript-eslint/no-unused-vars
     } catch (e) {
       navigate(`/case/${caseId}/egress-connect/error`, {
@@ -63,7 +63,9 @@ const EgressConnectConfirmationPage: React.FC = () => {
   };
   return (
     <div className={styles.confirmationWrapper}>
-      <BackLink to={backLinkUrl}>Back</BackLink>
+      <BackLink to={backLinkUrl} state={{ isRouteValid: true }}>
+        Back
+      </BackLink>
       <PageContentWrapper>
         <form onSubmit={handleSubmit}>
           <Radios
--- a/ui-spa/src/components/egress-connect/EgressConnectConfirmationPage.tsx
+++ b/ui-spa/src/components/egress-connect/EgressConnectConfirmationPage.tsx
@@ -33,8 +33,8 @@ const EgressConnectConfirmationPage: React.FC = () => {
 
   const handleSubmit = async (e: React.FormEvent<HTMLFormElement>) => {
     e.preventDefault();
-    if (formValue === "no" && state?.backLinkUrl) {
-      navigate(state.backLinkUrl, {
+    if (formValue === "no" && backLinkUrl) {
+      navigate(backLinkUrl, {
         state: {
           isRouteValid: true,
           searchQueryString,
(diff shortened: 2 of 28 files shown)


ui-spa/src/components/netapp-connect/index.tsx
```diff
@@ -60,92 +59,23 @@ const NetAppPage = () => {
   };
 
   const handleConnectFolder = (path: string) => {
```
HasanCPS, line 64: ### Preserve the current NetApp folder when returning from confirmation

```tsx
const handleConnectFolder = (path: string) => {
  navigate(`/case/${caseId}/netapp-connect/confirmation`, {
    state: {
      isRouteValid: true,
      operationName,
      caseId,
      searchQueryString,
      backLinkUrl: `/case/${caseId}/netapp-connect?operation-name=${operationName}`,
      selectedWorkspace: {
        folderPath: path,
      },
    },
  });
};
```

The confirmation page now lives on a separate route. When the user clicks **Back** or chooses **No**, `NetAppPage` is mounted again and `rootFolderPath` starts from `""`:

```tsx
const [rootFolderPath, setRootFolderPath] = useState("");
```

That means a valid flow like `Home > thunderstrike > folder-0 -> Connect -> Back` returns the user to `Home`, not to `Home > thunderstrike`. The updated Playwright expectation now accepts this reset, but it is a regression from the previous single-component flow where the selected folder context was still in memory.

The fix should carry the current folder path through the confirmation route, or encode it in the URL, and initialise `rootFolderPath` from that value when returning to `/netapp-connect`.

Related return points:

```tsx
// ui-spa/src/components/netapp-connect/NetAppConnectConfirmationPage.tsx:35
navigate(backLinkUrl, {
  state: {
    isRouteValid: true,
    searchQueryString,
  },
});

// ui-spa/src/components/netapp-connect/NetAppConnectConfirmationPage.tsx:66
<BackLink
  to={backLinkUrl}
  state={{ isRouteValid: true, searchQueryString }}
>
```

ui-spa/src/components/egress-connect/index.tsx
```diff
@@ -81,95 +60,42 @@ const EgressPage = () => {
       return;
     }
     setFormDataErrorText("");
-    setSearchParams({ "workspace-name": formValue });
+    navigate(`/case/${caseId}/egress-connect?workspace-name=${formValue}`, {
```
HasanCPS: ### URL-encode manual Egress workspace searches

```tsx
navigate(`/case/${caseId}/egress-connect?workspace-name=${formValue}`, {
  state: {
    isRouteValid: true,
    searchQueryString,
    isNetAppConnected,
  },
});
```

This replaces the previous `setSearchParams({ "workspace-name": formValue })` behaviour with manual string interpolation. `setSearchParams` encoded the user-entered value; this version does not.

For example, searching for `Smith & Co` navigates to:

```text
/case/12/egress-connect?workspace-name=Smith & Co
```

`useSearchParams().get("workspace-name")` will read only `Smith ` because `& Co` is parsed as another query parameter. A `+` in the workspace name is also decoded as a space. The API search then runs with the wrong workspace name, and the same bad value is reused in the confirmation back link.

Use `URLSearchParams` / `createSearchParams`, or pass a location object to `navigate`, so `workspace-name` is encoded before it is placed in the URL.

Example:

```tsx
const search = new URLSearchParams({ "workspace-name": formValue }).toString();

navigate(`/case/${caseId}/egress-connect?${search}`, {
  state: {
    isRouteValid: true,
    searchQueryString,
    isNetAppConnected,
  },
});
```

renjithabby: fixed

ui-spa/src/components/netapp-connect/index.tsx
renjithabby: fixed

ui-spa/src/components/egress-connect/EgressConnectConfirmationPage.tsx
```diff
@@ -1,25 +1,91 @@
...
+        });
+
+        if (!isNetAppConnected) {
+          const payload: SharedDriveConnectRouteState = {
+            isRouteValid: true,
+            searchQueryString: searchQueryString,
+            netappRootFolderPath: "",
+          };
+          navigate(
+            `/case/${caseId}/netapp-connect?operation-name=${selectedWorkspace.name}`,
```
HasanCPS: ### Encode operation and workspace names before putting them in route query strings

The protected-route refactor still builds several connect/reconnect URLs by interpolating raw operation or workspace names into the query string. That breaks when the name contains URL-reserved characters.

For example, this code:

```tsx
`/case/${caseId}/netapp-connect?operation-name=${selectedWorkspace.name}`
```

will produce a broken query string for a name such as:

```text
R v Smith & Jones
```

The URL becomes:

```text
/case/12/netapp-connect?operation-name=R v Smith & Jones
```

`useSearchParams().get("operation-name")` will then read `R v Smith ` and treat ` Jones` as a separate query part. In practice, the NetApp folder search can ask the API for the wrong operation and show the wrong or empty folder list after the user connects Egress.

Same pattern appears in these changed paths:

```text
ui-spa/src/components/search-result-page/SearchResults.tsx:29
ui-spa/src/components/search-result-page/SearchResults.tsx:45
ui-spa/src/components/netapp-connect/index.tsx:70
ui-spa/src/components/case-management/index.tsx:46
ui-spa/src/components/case-management/index.tsx:59
ui-spa/src/components/case-management/transfer-materials/index.tsx:338
ui-spa/src/components/case-management/transfer-materials/index.tsx:363
ui-spa/src/components/case-management/transfer-materials/MetaDataErrorPage.tsx:30
ui-spa/src/components/case-management/transfer-materials/MetaDataErrorPage.tsx:43
```

Use `URLSearchParams` everywhere these URLs are built, the same way `EgressPage` now does for the search form.

```tsx
const search = new URLSearchParams({
  "operation-name": selectedWorkspace.name,
});

navigate(`/case/${caseId}/netapp-connect?${search}`, {
  state: payload,
});
```

For Egress links:

```tsx
const search = new URLSearchParams({
  "workspace-name": operationName ?? "",
});

to={`/case/${data.caseId}/egress-connect?${search}`}
```

renjithabby: fixed

ui-spa/src/common/utils/getUrlSearchParam.ts
```diff
@@ -0,0 +1,6 @@
+export const getUrlSearchParam = (
+  property: string,
+  value: string | null | undefined,
+) => {
+  return new URLSearchParams({ [property]: value ?? "`" });
```
HasanCPS: ### Do not encode missing route query values as a backtick

The new helper turns `null` or `undefined` into a real query value:

```ts
return new URLSearchParams({ [property]: value ?? "`" });
```

That means a missing operation/workspace name becomes:

```text
workspace-name=%60
operation-name=%60
```

The connect pages treat any non-empty query value as valid. For example, `EgressPage` reads `workspace-name`, skips the empty-value validation, and searches Egress for the literal backtick value. That can happen if a caller passes through a missing `operation-name`, such as from `MetaDataErrorPage` where `searchParams.get("operation-name")` can return `null`.

Locations that can pass nullable values into this helper:

```text
ui-spa/src/components/case-management/transfer-materials/MetaDataErrorPage.tsx:31
ui-spa/src/components/case-management/transfer-materials/MetaDataErrorPage.tsx:44
```

I would make callers handle missing values before navigation, or make the helper preserve the existing empty-value behaviour:

```ts
return new URLSearchParams({ [property]: value ?? "" });
```

Then update the helper tests so `null` and `undefined` do not expect `%60`:

```text
ui-spa/src/common/utils/getUrlSearchParam.test.ts:9
ui-spa/src/common/utils/getUrlSearchParam.test.ts:14
```

renjithabby: fixed

HasanCPS (approved):
LGTM!

PR #436, opened 2026-05-15
Split the share resource and enforce registercase skip at entry point

kmshrajcps:
FCT2-17942: split Egress upload into disjoint copy-* and move-* file
   sets so the Move journey no longer consumes the Copy journey's source.
   Setup-EgressWorkspaceAndUpload.ps1 gains -MoveFileCount and emits new
   egressCopyFile* / egressMoveFile* variable sets; legacy egressFile*
   remain as aliases of the Copy set. Each E2E folder's file-lookup
   pre-request detects which journey it is in (via
   pm.execution.location.path) and routes to the matching set.

   FCT2-17943: enforce registerCase=false / pre-existing caseId at the
   request level. All six Register Case requests self-guard via
   pm.execution.skipRequest(); the previous setNextRequest-based upstream
   skips left the case-create API reachable from any entry point that
   did not pass through the guard request (single-folder Runner pass,
   manual Send, folders whose upstream skip never fired).

Capture23
Capture22
Capture21


e2e/postman/LCCUserJourneyTests.postman_collection.json
```diff
@@ -3637,6 +3741,23 @@
 									"console.log('=== List Files in Folder (Step 7b) ===');",
 									"console.log('Files found:', files.length);",
 									"",
+									"// === Journey selector (Copy vs Move source files) ===",
+									"// Source files are uploaded as disjoint copy-*/move-* sets so the two",
+									"// journeys don't share state. Pick the right set for this folder before",
+									"// the existing lookup runs.",
+									"const _journeyPath = (pm.execution && pm.execution.location && pm.execution.location.path) || [];",
```
HasanCPS: The new journey selector reads:

```
  const _journeyPath = (pm.execution && pm.execution.location && pm.execution.location.path) || [];
  const _inMove = _journeyPath.some(p => /Egress to NetApp Move/i.test(String(p)));
```

But Postman documents `pm.execution.location` itself as the array path, not` pm.execution.location.path`. So `_journeyPath` becomes `[]`, _`inMove` is always `false`, and the Move folder uses `egressCopy*` variables instead of `egressMove*`.

 Example impact:

  Run all:
  1. Copy journey uses copy-*.txt
  2. Move journey should use move-*.txt
  3. Actual behaviour: Move journey also uses copy-*.txt

  Fix shape:

  ```
const _journeyPath = (pm.execution && pm.execution.location) || [];
const _inMove = _journeyPath.some(p => /Egress to NetApp Move/i.test(String(p)));
```

File: e2e/postman/Run-E2E-Tests.ps1 [hunk 2]
```diff
@@ -147,6 +148,7 @@ $Config = @{
 $missingConfig = @()
 if (-not $Config.TenantId) { $missingConfig += "LCC_TENANT_ID" }
 if (-not $Config.LccApiId) { $missingConfig += "LCC_API_ID" }
+if (-not $Config.RegisterCaseClientId) { $missingConfig += "LCC_REGISTER_CASE_CLIENT_ID" }
```
HasanCPS: The script validates `LCC_REGISTER_CASE_CLIENT_ID` unconditionally:

 ` if (-not $Config.RegisterCaseClientId) { $missingConfig += "LCC_REGISTER_CASE_CLIENT_ID" }`

  But default mode explicitly uses an existing case/workspace and does not need to register a case:

  ```
if (-not $RegisterCase) {
      if (-not $Config.DefaultCaseId) { ... }
  }
```

 So a valid default-mode run can now fail at startup just because register-case credentials are absent.

 Also, when -RegisterCase is used, the new `LCC_DDEI_ACCESS_KEY_REGCASE` is required by the collection but is not validated here.

Suggested shape:

```
  if ($RegisterCase) {
      if (-not $Config.RegisterCaseClientId) { $missingConfig += "LCC_REGISTER_CASE_CLIENT_ID" }
      if (-not $Config.DdeiAccessKeyRegCase) { $missingConfig += "LCC_DDEI_ACCESS_KEY_REGCASE" }
  }
```

File: e2e/postman/LCCUserJourneyTests.postman_collection.json [hunk 3]
```diff
@@ -196,7 +196,7 @@
 							},
 							{
 								"key": "x-functions-key",
-								"value": "{{ddeiAccessKey}}"
+								"value": "{{ddeiAccessKeyRegCase}}"
```
HasanCPS: The top-level `1. Authentication > 1.2 Get CMS Auth (DDEI) `request now sends:

```
  {
    "key": "x-functions-key",
    "value": "{{ddeiAccessKeyRegCase}}"
  }
```

That request uses the LCC token from `lccApiId`, and its own description still says it requires `ddeiAccessKey`. The register-case key should only be used by the register-case auth path. This will break anyone running the generic authentication folder with the normal LCC DDEI key.

e2e/postman/Run-E2E-Tests.ps1
HasanCPS: ###  Parse the upload JSON by shape, not by first property

The orchestrator only tries to parse the upload summary JSON when the line starts with `{"workspaceId"`:

```powershell
if ($line -match '^\s*\{"workspaceId"') {
    $jsonData = $line.Trim() | ConvertFrom-Json
    ...
}
```

That is a fragile contract. The upload script builds the JSON from a normal PowerShell hashtable at `e2e/postman/Setup-EgressWorkspaceAndUpload.ps1:865`, so the parser should not rely on `workspaceId` being the first property. If PowerShell emits the same object with another property first, for example:

```json
{"workspaceName":"AUTOMATION-TESTING","workspaceId":"...","files":[...]}
```

the JSON parse is skipped and the script falls back to regex parsing. That fallback only fills the legacy `egressFile*` values. It does not populate `egressMoveFileNames` or `egressMoveFileIds`.

That matters because the move folder now relies on those variables at `e2e/postman/LCCUserJourneyTests.postman_collection.json:5611`. If they are missing, the move journey keeps the legacy copy aliases and can move the copy source instead of the dedicated move source.

Suggested fix:

```powershell
foreach ($line in $uploadLines) {
    try {
        $jsonData = $line.Trim() | ConvertFrom-Json -ErrorAction Stop

        if ($jsonData.workspaceId -and $jsonData.files) {
            # existing JSON mapping here
            break
        }
    }
    catch {
        # Ignore non-JSON log lines
    }
}
```

I would also either output the summary as an `[ordered]` object from `Setup-EgressWorkspaceAndUpload.ps1`, or extend the fallback regex path to parse `egressCopy*` and `egressMove*` variables as well.

HasanCPS: ### Copy-only and NetApp-to-Egress runs still upload an unused move file

`Run-E2E-Tests.ps1` passes only `FileCount` to the upload script:

```powershell
$uploadArgs["FileCount"] = $FileCount
```

But `Setup-EgressWorkspaceAndUpload.ps1:70` now defaults `MoveFileCount` to `1`. That means commands such as:

```powershell
.\Run-E2E-Tests.ps1 -TestsToRun copy -SizeGB 1
```

still upload one copy file and one move file, even though the move folder is not going to run. For large-file runs this can double setup time and storage use without adding coverage.

Suggested fix: have the orchestrator pass `MoveFileCount = 0` unless `TestsToRun` is `all` or `move`. If NetApp-to-Egress does not need an uploaded Egress source file, consider avoiding both copy and move uploads for that mode, or make that choice explicit in the script output.

e2e/postman/LCCUserJourneyTests.postman_collection.json
HasanCPS, line 2838: ### Default mode still sends the register-case Azure token request

The script validation now treats `LCC_REGISTER_CASE_CLIENT_ID` as required only for `-RegisterCase` at `e2e/postman/Run-E2E-Tests.ps1:164`, but each E2E folder still starts by sending the first Azure token request with `{{registerCaseClientId}}`:

```json
{
  "key": "client_id",
  "value": "{{registerCaseClientId}}"
}
```

The default-mode branch that skips case registration runs in the post-response script, so the request has already gone to Azure before the skip logic runs. If `LCC_REGISTER_CASE_CLIENT_ID` is empty in default mode, this will make a noisy failed token request before moving on to `1b. Get LCC API Token`.

This may not fail Newman because the default branch does not assert the first response status, but it makes the "default mode does not exercise these" assumption misleading.

Suggested fix: move the default-mode skip into a pre-request script so the register-case token call is not sent.

e2e/postman/LCCUserJourneyTests.postman_collection.json
```diff
@@ -2659,6 +2743,29 @@
...
+									"// Default mode: skip the register-case Azure token call entirely.",
+									"// This request uses {{registerCaseClientId}}, which is only required",
+									"// when -RegisterCase is passed. Without this guard the request still",
+									"// fires against Azure with an empty/invalid client_id and 400s before",
+									"// the post-response skip can run.",
+									"if (pm.variables.get('registerCase') !== 'true') {",
+									"    pm.environment.set('caseId', pm.variables.get('defaultCaseId'));",
+									"    pm.environment.set('caseUrn', pm.variables.get('defaultCaseUrn'));",
+									"    console.log('[skip] registerCase!=true; using defaultCaseId:', pm.variables.get('defaultCaseId'));",
+									"    if (typeof pm.execution.skipRequest === 'function') pm.execution.skipRequest();",
```
HasanCPS: ### Default-mode Newman runs still fall into the register-case auth request

Location:

```text
e2e/postman/LCCUserJourneyTests.postman_collection.json:2759
e2e/postman/LCCUserJourneyTests.postman_collection.json:4642
e2e/postman/LCCUserJourneyTests.postman_collection.json:6529
```

The new default-mode guard is trying to skip the register-case Azure token request and jump straight to `1b. Get LCC API Token`:

```javascript
if (pm.variables.get('registerCase') !== 'true') {
    pm.environment.set('caseId', pm.variables.get('defaultCaseId'));
    pm.environment.set('caseUrn', pm.variables.get('defaultCaseUrn'));
    if (typeof pm.execution.skipRequest === 'function') pm.execution.skipRequest();
    pm.execution.setNextRequest('1b. Get LCC API Token');
    return;
}
```

The problem is the order. `pm.execution.skipRequest()` stops the rest of the pre-request script and moves to the next request in folder order. That means the `setNextRequest('1b. Get LCC API Token')` line after it is not reliable.

In default mode, the next request in all three E2E folders is `1a. Get CMS Auth for Register Case`. So the run can still go there with no register-case `accessToken`, and default mode also no longer requires `ddeiAccessKeyRegCase`. That breaks the flow before it reaches the LCC API token request.

Example failure path:

```text
E2E: Egress to NetApp Copy
1. Get Azure AD Token -> skipped
1a. Get CMS Auth for Register Case -> runs next with missing register-case auth values
1b. Get LCC API Token -> not reached
```

This affects:

```text
E2E: Egress to NetApp Copy
E2E: Egress to NetApp Move
E2E: Netapp to Egress Copy
```

I would move `setNextRequest('1b. Get LCC API Token')` before `skipRequest()`, then test one default-mode folder through Newman. If that still does not jump reliably, the safer fix is to avoid using `skipRequest()` for this branch and split the default/register-case startup sequence more explicitly.

e2e/postman/Setup-EgressWorkspaceAndUpload.ps1
```diff
@@ -66,6 +66,9 @@ param(
     [Parameter(Mandatory=$false)]
     [int]$FileCount = 1,
 
+    [Parameter(Mandatory=$false)]
+    [int]$MoveFileCount = 1,
```
HasanCPS: ### Standalone upload helper now uploads a move file by default

`MoveFileCount` defaults to `1`:

```powershell
[Parameter(Mandatory=$false)]
[int]$MoveFileCount = 1
```

That changes the standalone helper's documented behaviour. A caller running the existing examples now gets two uploads instead of one:

```powershell
.\Setup-EgressWorkspaceAndUpload.ps1 -SizeMB 100
```

Before this PR, that command uploaded one file. With this change it uploads one copy file and one move file. Likewise, `-FileCount 3` now uploads six files unless the caller knows to pass `-MoveFileCount 0`.

`Run-E2E-Tests.ps1` already passes `MoveFileCount` explicitly based on `-TestsToRun`, so the helper itself can default `MoveFileCount` to `0` and still keep the new split-source behaviour for orchestrated runs.

HasanCPS (approved):
LGTM!

PR #438, opened 2026-05-18
Feature/fct2 16467 automated net app folder provisioning from template

NeilFoubister:


backend/CPS.ComplexCases.API/Functions/ProvisionNetAppFolders.cs
```diff
@@ -0,0 +1,134 @@
...
+            CaseId = caseId,
+            Urn = cmsResponse.Urn,
+            TemplateName = provisionFoldersRequest.Value.TemplateFolderPath,
+            DestinationFolderPath = folderPathName,
+            BucketName = securityGroups.First().BucketName,
+            BearerToken = context.BearerToken,
+            UserName = context.Username
+        };
+
+        await _transferClient.ProvisionNetAppFoldersAsync(provisionNetAppFoldersRequest, context.CorrelationId);
```
HasanCPS: ### Do not create NetApp connection metadata from a schedule-only or failed provisioning response

The gateway calls the File Transfer provisioning endpoint, but it ignores the returned `HttpResponseMessage` and immediately writes the NetApp connection metadata.

```csharp
await _transferClient.ProvisionNetAppFoldersAsync(provisionNetAppFoldersRequest, context.CorrelationId);

await _caseMetadataService.CreateNetAppConnectionAsync(new CreateNetAppConnectionDto
{
    CaseId = caseId,
    NetAppFolderPath = folderPathName,
    OperationName = cmsResponse.OperationName ?? cmsResponse.LeadDefendantSurname ?? caseName
});
```

That leaves a bad state if File Transfer rejects the request. For example, the File Transfer endpoint returns `400 BadRequest` when the template folder has no files:

```csharp
return new BadRequestObjectResult($"No files found in the specified template folder: {request.TemplateName}");
```

but the gateway will still create the NetApp connection and return `200 OK`. The case then looks connected to a folder that was never provisioned, and a retry is blocked by the "already has a NetApp folder" check.

There is a second part to this: even a `200 OK` from File Transfer only means the durable orchestration was scheduled. The endpoint returns before the copy finishes:

```csharp
return new OkObjectResult(templateObjects); // Return immediately while the provisioning continues asynchronously
```

So if the orchestration later fails or only partially copies the template, the gateway has already marked the case as connected.

At minimum, handle the immediate non-success response before writing metadata:

```csharp
var response = await _transferClient.ProvisionNetAppFoldersAsync(
    provisionNetAppFoldersRequest,
    context.CorrelationId);

if (!response.IsSuccessStatusCode)
{
    return await response.ToActionResult();
}
```

That still leaves the async failure case. The safer shape is to create the connection metadata only after provisioning completes successfully, or return `202 Accepted` with a provisioning status and keep the case unconnected or pending until the durable workflow finishes.

backend/CPS.ComplexCases.API/Validators/Requests/ProvisionNetAppFoldersRequestValidator.cs
```diff
@@ -0,0 +1,15 @@
+using CPS.ComplexCases.NetApp.Models.Dto;
+using FluentValidation;
+
+namespace CPS.ComplexCases.API.Validators.Requests;
+
+public class ProvisionNetAppFoldersRequestValidator : AbstractValidator<ProvisionNetAppFoldersDto>
+{
+    public ProvisionNetAppFoldersRequestValidator()
+    {
+        RuleFor(x => x.TemplateFolderPath)
```
HasanCPS, line 10: ### Restrict template paths so callers cannot copy arbitrary NetApp folders

The public API accepts a raw `templateFolderPath` and only blocks `..` and a leading slash.

```csharp
RuleFor(x => x.TemplateFolderPath)
    .NotEmpty().WithMessage("Path is required.")
    .Must(path => !path.Contains("..")).WithMessage("Path cannot contain '..' to navigate up directories.")
    .Must(path => !path.StartsWith('/')).WithMessage("Path cannot start with a '/'.");
```

The File Transfer endpoint then lists and copies whatever prefix it receives:

```csharp
var files = await ListFilesInFolder(
    request.TemplateName,
    request.BearerToken ?? throw new ArgumentNullException(nameof(request.BearerToken), "Bearer token cannot be null."),
    request.BucketName ?? throw new ArgumentNullException(nameof(request.BucketName), "Bucket name cannot be null."));
```

That means a caller could send another folder in the same bucket instead of a template:

```json
{
  "templateFolderPath": "AnotherCase-45AA2098221/"
}
```

If the caller has access to that bucket, the provisioning flow will copy those objects into the new case folder. This bypasses the normal Manage Materials source-path checks, because the source no longer has to be inside the current case folder.

The validator should only allow known template locations, ideally from configuration or template IDs rather than arbitrary object prefixes. If the path approach stays, normalise and enforce something like:

```csharp
RuleFor(x => x.TemplateFolderPath)
    .NotEmpty()
    .Must(path => path.StartsWith("_templates/", StringComparison.OrdinalIgnoreCase))
    .Must(path => path.EndsWith('/'));
```

The trailing slash matters too. Without it, a prefix such as `_templates/appeal` can also match `_templates/appeal2/...`.

backend/CPS.ComplexCases.API/Functions/ProvisionNetAppFolders.cs
```diff
@@ -0,0 +1,134 @@
...
+
+        await _caseMetadataService.CreateNetAppConnectionAsync(new CreateNetAppConnectionDto
+        {
+            CaseId = caseId,
+            NetAppFolderPath = folderPathName,
+            OperationName = cmsResponse.OperationName ?? cmsResponse.LeadDefendantSurname ?? caseName
```
HasanCPS: ###  Preserve the existing case-name fallback when storing `OperationName`

The folder path is generated through `CaseNamingService`, which treats a blank operation name as missing and falls back to the lead defendant surname.

The metadata write does not use the same rule:

```csharp
OperationName = cmsResponse.OperationName ?? cmsResponse.LeadDefendantSurname ?? caseName
```

If CMS returns `OperationName = ""`, the new NetApp connection stores an empty operation name even though the folder name was generated from the lead defendant. The normal `CreateNetAppConnection` path would reject an empty operation name through `CreateNetAppConnectionValidator`, but this new path bypasses that validator and writes metadata directly.

Use the same whitespace-aware fallback here, or expose it from the naming service:

```csharp
OperationName = !string.IsNullOrWhiteSpace(cmsResponse.OperationName)
    ? cmsResponse.OperationName
    : !string.IsNullOrWhiteSpace(cmsResponse.LeadDefendantSurname)
        ? cmsResponse.LeadDefendantSurname
        : caseName;
```

backend/CPS.ComplexCases.API/Functions/ProvisionNetAppFolders.cs
```diff
@@ -0,0 +1,209 @@
...
+
+        var responseBody = await response.Content.ReadAsStringAsync();
+        var parsedResponse = JsonSerializer.Deserialize<JsonElement>(responseBody);
+        var transferId = parsedResponse.GetProperty("id").GetString() ?? string.Empty;
+
+        var finalStatus = response.StatusCode == HttpStatusCode.Accepted
+            ? await PollForTerminalStatusAsync(transferId, context.CorrelationId)
+            : await GetTransferStatusAsync(transferId, context.CorrelationId);
+
+        if (finalStatus == null || finalStatus == "Failed")
```
HasanCPS: Can we only write the case metadata when the copy has fully completed? Right now the gateway blocks `null` and `Failed`, but `PartiallyCompleted` only logs a warning and then still creates the NetApp connection record.

On the `202 Accepted` path, polling waits for a terminal status. On the immediate-response path, the gateway calls `GetTransferStatusAsync` once, so any non-null status other than `Failed` can fall through to `CreateNetAppConnectionAsync`.

```csharp
if (finalStatus == null || finalStatus == "Failed")
{
    return new ObjectResult("NetApp folder provisioning failed.")
    {
        StatusCode = (int)HttpStatusCode.InternalServerError
    };
}

if (finalStatus == "PartiallyCompleted")
{
    _logger.LogWarning("NetApp folder provisioning partially completed...");
}

await _caseMetadataService.CreateNetAppConnectionAsync(...);
```

That can leave the case looking provisioned even when some template files or folder markers did not copy. A retry is then blocked because `NetappFolderPath` has already been set.

I would only write metadata when `finalStatus == "Completed"`. `PartiallyCompleted`, `Failed`, `Initiated`, `InProgress`, and unknown statuses should not link the case to the new folder.

backend/CPS.ComplexCases.FileTransfer.API/Functions/ProvisionNetAppFolders.cs
```diff
@@ -0,0 +1,231 @@
...
+            }
+
+            // Recursively process all subdirectories
+            if (response.Data.FolderData.Any())
+            {
+                foreach (var folder in response.Data.FolderData)
+                {
+                    if (!string.IsNullOrEmpty(folder.Path))
+                    {
+                        filesForTransfer.Add(new FileTransferInfo
```
HasanCPS: Can we split folder entries from real files before the transfer fan-out? The ticket expects blank templates to work. In practice that means a template may contain empty subfolders as well as small placeholder files.

`ListFilesInFolder` currently adds `FolderData` entries to the same list as real files. The orchestrator then uses those entries to create the destination folders, but it still passes the full list into `FanOutTransferFilesAsync`.

```csharp
filesForTransfer.Add(new FileTransferInfo
{
    SourcePath = folder.Path
});
```

```csharp
var folders = input.Files.Where(f => f.IsFolder).ToList();
if (folders.Count > 0)
    await CreateFolderStructure(context, input, folders);

var allResults = await FanOutTransferFilesAsync(context, input.Files, BuildPayload, entityId, telemetryEvent);
```

If `FolderData` is just an S3 common prefix, `TransferFile` will try to `GetObject` for something that is not a real object. If it is a zero-byte folder marker, the marker can be copied as a file as well as being handled by `CreateDestinationFolder`. Either way, folder entries are being handled twice.

Examples:

This should work and create the same empty folder structure:

```text
Template:
_templates/charged/
_templates/charged/correspondence/
_templates/charged/exhibits/

Expected destination:
OperationName-12AB3456/
OperationName-12AB3456/correspondence/
OperationName-12AB3456/exhibits/
```

This should also work. The folder structure should be created, but only the real file should go through `TransferFile`:

```text
Template:
_templates/charged/
_templates/charged/correspondence/
_templates/charged/readme.txt

Expected behaviour:
- create OperationName-12AB3456/
- create OperationName-12AB3456/correspondence/
- copy _templates/charged/readme.txt to OperationName-12AB3456/readme.txt
- do not try to copy _templates/charged/ or _templates/charged/correspondence/ as files
```

Small test ask: can we add an assertion for the final destination key for a normal `.txt` file? I have seen this kind of issue show up as an extra extension, so it is worth locking down.

```text
Template file:
_templates/charged/note.txt

Expected destination:
OperationName-12AB3456/note.txt

Should not become:
OperationName-12AB3456/note.txt.txt
```

The normal path should keep the relative filename as-is:

```csharp
RelativePath = file.SourcePath.RemovePathPrefix(request.TemplateName)
DestinationFileName = obj.RelativePath!
```

and then combine it with the destination folder:

```csharp
var objectName = Path.Combine(destinationPath, relativePath ?? string.Empty).Replace('\\', '/');
```

I would cover this with a focused unit test around the final object key. If the code can take both single PUT and multipart paths, it is worth asserting both produce `OperationName-12AB3456/note.txt`.

This should still be rejected, because there is nothing to copy or recreate:

```text
Template:
_templates/charged/

NetApp listing:
no FileData
no FolderData

Expected behaviour:
return 400/404 and do not schedule the copy orchestration
```

I would keep two lists: one for folders to create, and one for real file objects to transfer. At minimum, filter the transfer fan-out to `input.Files.Where(f => !f.IsFolder)` after creating the folder structure. If the template listing has no files and no folder entries, it should still be treated as missing/invalid.

backend/CPS.ComplexCases.API/Validators/Requests/ProvisionNetAppFoldersRequestValidator.cs
```diff
@@ -0,0 +1,16 @@
...
```
HasanCPS, line 10: Can we make this validator null-safe? It starts with `NotEmpty()`, but the later `Must(...)` calls still dereference `path`. If the client sends `{}` or `"templateFolderPath": null`, the validator can throw before returning the intended `"Path is required."` error.

```csharp
RuleFor(x => x.TemplateFolderPath)
    .NotEmpty().WithMessage("Path is required.")
    .Must(path => path.StartsWith("_templates/", StringComparison.OrdinalIgnoreCase))
    .Must(path => path.EndsWith('/'))
    .Must(path => !path.Contains("..")).WithMessage("Path cannot contain '..' to navigate up directories.");
```

Use `Cascade(CascadeMode.Stop)` or null-safe predicates:

```csharp
RuleFor(x => x.TemplateFolderPath)
    .Cascade(CascadeMode.Stop)
    .NotEmpty().WithMessage("Path is required.")
    .Must(path => path.StartsWith("_templates/", StringComparison.OrdinalIgnoreCase))
    .WithMessage("Path must be under _templates/.")
    .Must(path => path.EndsWith('/'))
    .WithMessage("Path must end with '/'.")
    .Must(path => !path.Contains(".."))
    .WithMessage("Path cannot contain '..' to navigate up directories.");

File: backend/CPS.ComplexCases.FileTransfer.API/Functions/ProvisionNetAppFolders.cs [hunk 7]
```diff
@@ -0,0 +1,231 @@
...
+        }
+        else
+        {
+            _logger.LogWarning("No files found in the specified template folder. CorrelationId: {CorrelationId}, TemplateName: {TemplateName}",
+                correlationId, request.TemplateName);
+
+            return new BadRequestObjectResult($"No files found in the specified template folder: {request.TemplateName}");
+        }
+
+        foreach (var obj in templateObjects)
```
HasanCPS, line 97: Can we add a destination check before scheduling the copy? Even though this is provisioning a new case folder, the destination can still already exist after a previous timed-out or partially completed attempt.

At the moment this endpoint builds `CopyFileItem`s straight into `request.DestinationFolderPath`, but I cannot see anything that checks whether the target folder, or any of the final object keys, already exists. The shared `TransferFile` activity only checks destination conflicts for `EgressToNetApp`. This new path is `NetAppToNetApp`, so if the folder is already there, the retry can write into it again.

```csharp
foreach (var obj in templateObjects)
{
    copyFileItems.Add(new CopyFileItem
    {
        SourceKey = obj.SourcePath,
        DestinationPrefix = request.DestinationFolderPath,
        DestinationFileName = obj.RelativePath!,
    });
}
```

Example:

```text
First attempt:
copies some files to Smith-12AB3456/
times out before metadata is written

Retry:
case still looks unlinked
provisioning runs again into Smith-12AB3456/
```

The ticket also says the operation should fail cleanly if the target folder name already exists, so I think this should be handled before the durable workflow starts.

I would list the destination prefix with `MaxKeys = 1` and return `409` if anything already exists. It would also be worth checking the computed destination keys before scheduling the copy.

backend/CPS.ComplexCases.API/Functions/ProvisionNetAppFolders.cs
```diff
@@ -0,0 +1,210 @@
...
+        };
+
+        var response = await _transferClient.ProvisionNetAppFoldersAsync(provisionNetAppFoldersRequest, context.CorrelationId);
+
+        if (!response.IsSuccessStatusCode)
+        {
+            _logger.LogError("Failed to provision NetApp folders for caseId: {CaseId}. StatusCode: {StatusCode}, Response: {Response}",
+                caseId, response.StatusCode, await response.Content.ReadAsStringAsync());
+
+            return response.StatusCode switch
```
HasanCPS, line 122: ###  Return `409` from the Gateway provision endpoint when the destination already exists

Can we pass through the FileTransfer API conflict response here? The FileTransfer endpoint now correctly returns `ConflictObjectResult` when the destination folder already exists, but the public API wrapper does not map `HttpStatusCode.Conflict`. It falls into the default case and returns `500`.

Current gateway API mapping:

```csharp
return response.StatusCode switch
{
    HttpStatusCode.BadRequest => new BadRequestObjectResult("Invalid request to provision NetApp folders."),
    HttpStatusCode.Unauthorized => new UnauthorizedObjectResult("Unauthorized to provision NetApp folders."),
    HttpStatusCode.Forbidden => new ForbidResult("Forbidden from provisioning NetApp folders."),
    HttpStatusCode.NotFound => new NotFoundObjectResult("Template folder not found for provisioning NetApp folders."),
    _ => new ObjectResult("An error occurred while provisioning NetApp folders.") { StatusCode = (int)HttpStatusCode.InternalServerError }
};
```

FileTransfer now returns this when the destination exists:

```csharp
return new ConflictObjectResult(
    $"Destination folder '{request.DestinationFolderPath}' already exists. A previous provisioning attempt may have partially completed.");
```

Example:

```text
First attempt:
copies some files to Smith-12AB3456/
times out before metadata is written

Retry:
FileTransfer sees Smith-12AB3456/ already exists
FileTransfer returns 409
Public API turns that into 500
```

That means the previous destination-folder comment is only fixed inside FileTransfer. The user-facing endpoint still reports an internal error for a known conflict.

I would add:

```csharp
HttpStatusCode.Conflict => new ConflictObjectResult(await response.Content.ReadAsStringAsync()),
```

and a gateway unit test for the `409` case.

backend/CPS.ComplexCases.FileTransfer.API/Functions/ProvisionNetAppFolders.cs
```diff
@@ -0,0 +1,246 @@
...
+            });
+        }
+
+        var destinationCheckArg = _netAppArgFactory.CreateListObjectsInBucketArg(
+            request.BearerToken,
+            request.BucketName,
+            maxKeys: 1,
+            prefix: request.DestinationFolderPath);
+
+        var destinationCheck = await _netAppClient.ListObjectsInBucketAsync(destinationCheckArg);
```
HasanCPS, line 113: ###  Do not treat a failed destination preflight as an empty destination

Can we fail closed if the destination check returns `null`? The destination preflight now lists `request.DestinationFolderPath` with `MaxKeys = 1`, which is the right shape. But if NetApp listing fails, `NetAppClient.ListObjectsInBucketAsync` can return `null`, and this code currently treats that the same as "nothing exists there".

```csharp
var destinationCheck = await _netAppClient.ListObjectsInBucketAsync(destinationCheckArg);

if (destinationCheck?.Data.FileData.Any() == true || destinationCheck?.Data.FolderData.Any() == true)
{
    return new ConflictObjectResult(...);
}
```

`NetAppClient` returns `null` on list failures:

```csharp
catch (AmazonS3Exception ex)
{
    _logger.LogError(ex, "Failed to list objects in bucket {BucketName}.", arg.BucketName);
    return null;
}
```

There is also a test that locks in the risky behaviour:

```csharp
public async Task Run_WhenDestinationCheckReturnsNull_ProceedsToScheduleOrchestrator()
```

For this preflight, `null` is not proof that the destination is empty. It means the check failed. If we schedule the copy anyway, we are back to the original retry/idempotency risk.

I would return a `500`/`503` when the destination check is `null`, and change the test to assert that the orchestrator is not scheduled.

backend/CPS.ComplexCases.FileTransfer.API/Functions/ProvisionNetAppFolders.cs
```diff
@@ -0,0 +1,257 @@
...
+                bearerToken,
+                bucketName,
+                continuationToken,
+                1000,
+                path,
+                true);
+
+            var response = await _netAppClient.ListObjectsInBucketAsync(arg);
+
+            if (response == null)
```
HasanCPS, line 226: ###  Abort if any template listing call fails

I think we still need to fail closed while reading the template.

`ListObjectsInBucketAsync` returns `null` when the NetApp/S3 list call fails. In `ListFilesInFolder`, that `null` response is currently treated as "return whatever we have collected so far":

```csharp
var response = await _netAppClient.ListObjectsInBucketAsync(arg);

if (response == null)
{
    return filesForTransfer;
}
```

That is risky for template provisioning because the template is read recursively. For example:

```text
_templates/charged/
  Case Management/Document1.docx
  Evidence/Document2.docx
```

If the root listing succeeds, `Case Management/` lists successfully, but the recursive list for `Evidence/` returns `null`, we still schedule the copy with a partial set of template objects. The gateway can then link the case to a provisioned NetApp folder that is missing part of the agreed template.

This is different from an actually empty template. A failed list should come back as a provisioning failure, probably `503`, and should not schedule the orchestrator. I would add a test where a nested folder listing returns `null` and assert that no copy is scheduled.

Why I am calling this out: `NetAppClient.ListObjectsInBucketAsync` logs and returns `null` for `AmazonS3Exception` at `backend/CPS.ComplexCases.NetApp/Client/NetAppClient.cs:232`. The destination preflight now handles that correctly, but the template enumeration path still does not.

backend/CPS.ComplexCases.API/Functions/ProvisionNetAppFolders.cs
```diff
@@ -0,0 +1,212 @@
...
+                HttpStatusCode.ServiceUnavailable => new ObjectResult("NetApp service is currently unavailable.") { StatusCode = (int)HttpStatusCode.ServiceUnavailable },
+                _ => new ObjectResult("An error occurred while provisioning NetApp folders.") { StatusCode = (int)HttpStatusCode.InternalServerError }
+            };
+        }
+
+        var responseBody = await response.Content.ReadAsStringAsync();
+        var parsedResponse = JsonSerializer.Deserialize<JsonElement>(responseBody);
+        var transferId = parsedResponse.GetProperty("id").GetString() ?? string.Empty;
+
+        var finalStatus = response.StatusCode == HttpStatusCode.Accepted
```
HasanCPS: ###  Poll for the final entity status after a 200 response as well

The gateway now polls when File Transfer returns `202`, but the `200` path still reads transfer status once:

```csharp
var finalStatus = response.StatusCode == HttpStatusCode.Accepted
    ? await PollForTerminalStatusAsync(transferId, context.CorrelationId)
    : await GetTransferStatusAsync(transferId, context.CorrelationId);
```

I do not think one read is quite enough here. The File Transfer endpoint returns `200 OK` based on the durable orchestration runtime completing, but the actual transfer status is stored on the transfer entity. That final status is set through `FinalizeTransfer`, which sends an entity signal from an activity.

So a small template can finish inside the 25 second File Transfer wait window, return `200`, and then the gateway immediately reads the entity before the final `Completed`/`PartiallyCompleted` status has landed. If it sees `InProgress`, the gateway returns `400` and does not write case metadata, even though the folder copy may already have completed. A retry then hits the new destination-exists `409`, leaving a copied folder with no linked case metadata.

I would use the same polling path after both `200` and `202`, or change File Transfer to return the actual final entity status after it has been written.

Related locations:

```text
backend/CPS.ComplexCases.FileTransfer.API/Functions/ProvisionNetAppFolders.cs:186
backend/CPS.ComplexCases.FileTransfer.API/Functions/ProvisionNetAppFolders.cs:194
backend/CPS.ComplexCases.FileTransfer.API/Durable/Activity/FinalizeTransfer.cs:16
```

backend/CPS.ComplexCases.FileTransfer.API/Durable/Activity/CreateDestinationFolder.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            _logger.LogError("BucketName is required in CreateDestinationFolder activity.");
+            throw new ArgumentException("BucketName is required in CreateDestinationFolder activity.", nameof(payload));
+        }
+
+        var storageClient = _storageClientFactory.GetSourceClientForDirection(payload.TransferDirection);
+
+        try
+        {
+            await storageClient.CreateFolderAsync(payload.DestinationFolderPath, null, payload.BearerToken, payload.BucketName);
```
HasanCPS: ###  Fail the orchestration when an empty folder marker is not created

The folder creation activity ignores the boolean returned by `CreateFolderAsync`:

```csharp
await storageClient.CreateFolderAsync(payload.DestinationFolderPath, null, payload.BearerToken, payload.BucketName);
_logger.LogInformation("Successfully created destination folder: {BucketName}", payload.BucketName);
```

That return value matters. The NetApp folder PUT path returns `false` for non-success responses other than `401` or `403`, and `NetAppClient.CreateFolderCoreAsync` only logs the failure before returning `false`.

Example:

```text
Template contains:
_templates/charged/Appeals/

Destination should contain:
OperationName-12AB3456/Appeals/

If the PUT folder call returns a non-2xx response, CreateFolderAsync returns false.
CreateDestinationFolder still treats it as success.
The copy orchestration can complete and the public API can link the case to a folder that is missing that empty subfolder.
```

This matters for the CTD-style templates because a lot of the structure is empty folders, not just files. A failed marker creation should fail the activity so the orchestration moves to `Failed` and the public endpoint does not create the NetApp connection metadata.

Suggested fix:

```csharp
var created = await storageClient.CreateFolderAsync(
    payload.DestinationFolderPath,
    null,
    payload.BearerToken,
    payload.BucketName);

if (!created)
{
    throw new InvalidOperationException(
        $"Failed to create destination folder '{payload.DestinationFolderPath}' in bucket '{payload.BucketName}'.");
}
```

Add a unit test beside the existing exception tests:

```csharp
[Fact]
public async Task Run_WhenCreateFolderReturnsFalse_ThrowsInvalidOperationException()
{
    var payload = CreateValidPayload();

    _storageClientMock
        .Setup(c => c.CreateFolderAsync(
            It.IsAny<string>(),
            It.IsAny<string?>(),
            It.IsAny<string?>(),
            It.IsAny<string?>()))
        .ReturnsAsync(false);

    await Assert.ThrowsAsync<InvalidOperationException>(() => _activity.Run(payload));
}
```

HasanCPS (approved):
LGTM!

PR #439, opened 2026-05-19
FCT2-18461: NetApp changed to 'Shared Drive' in activity log when con…

NeilFoubister:
…necting/disconnecting to NetApp.


HasanCPS (approved):
LGTM!

PR #440, opened 2026-05-19
Task/fct2-18191 add playwright e2e tests to deployment pipeline

lilachdavis:
Includes some changes to the E2E test suite:
- Removed unnecessary CMS cookie from NetApp teardown
- Disambiguated the 'ClientId' variable
- Updated the required variables, as several were not necessary for the default route
- Obscured passwords from playwright report output
- Reduced uploaded files default size
- Updated navigation in CaseSearchPage.goto() to reduce CI long stalls during page load.


devops-pipelines/templates/e2e-playwright-test-job.yml
```diff
@@ -0,0 +1,77 @@
...
+        inputs:
+          version: "$(nodeVersion)"
+
+      - task: Npm@1
+        displayName: "Install NPM Dependencies"
+        inputs:
+          command: "ci"
+          workingDir: "$(workingDir)"
+
+      - script: node_modules/.bin/playwright install chromium
```
HasanCPS: ###  Keep Playwright's Linux dependency install in the Azure job

The latest change replaced:

```yaml
npx playwright install chromium --with-deps
```

with:

```yaml
node_modules/.bin/playwright install chromium
```

Using the local binary is fine, but dropping `--with-deps` means the job only downloads the browser. It no longer installs the Linux libraries Chromium needs. This template runs on an Azure pipeline agent and does not use the Playwright Docker image or any separate `install-deps` step, so a clean Linux agent can fail at browser launch with missing system dependencies.

Playwright's Azure Pipelines example keeps both pieces:

```yaml
- script: npx playwright install --with-deps
  displayName: 'Install Playwright browsers'
```

Source: <https://playwright.dev/docs/ci>

If Sonar is complaining about `npx`, keep the local binary but keep the dependency install:

```yaml
- script: node_modules/.bin/playwright install --with-deps chromium
  displayName: "Install Browsers"
  workingDirectory: $(workingDir)
```

If the LACC agents already have these OS packages baked in, I would still make that explicit in the template or agent image docs. Otherwise this job depends on hidden runner state.

devops-pipelines/templates/e2e-playwright-test-job.yml
```diff
@@ -0,0 +1,77 @@
...
+        displayName: "Install Browsers"
+        workingDirectory: $(workingDir)
+
+      - script: |
```
HasanCPS, line 44: ### Set `CI=true` for the Azure Playwright run

The new job runs the tests without setting `CI`:

```yaml
- script: |
    npm run e2e:existing-case
```

but Playwright CI behaviour in this repo depends on `process.env.CI`:

```ts
forbidOnly: !!process.env.CI,
retries: process.env.CI ? 2 : 0,
```

Location: `e2e/pw/playwright.config.ts:12`

Azure exposes `TF_BUILD=True` for build scripts, but that does not set `CI` by itself. Microsoft documents `TF_BUILD` as the Azure build indicator: <https://learn.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops>

So unless a variable group happens to define `CI`, this deployment gate runs with local Playwright behaviour: no retries, and `test.only` will not fail the pipeline. Playwright's own Azure example sets `CI: 'true'` on the test run step.

Add it explicitly:

```yaml
- script: |
    npm run e2e:existing-case
  workingDirectory: $(workingDir)
  displayName: "Run E2E Tests"
  env:
    CI: "true"
    BASE_URL: $(RedirectUrlLccUi)
```

or make the config understand Azure:

```ts
const isCI = !!process.env.CI || !!process.env.TF_BUILD;

export default defineConfig({
  forbidOnly: isCI,
  retries: isCI ? 2 : 0,
});
```

e2e/pw/fixtures/test-fixtures-default-large.ts
```diff
@@ -5,10 +5,10 @@ import { loadEnvConfig } from "../helpers/env-config";
 import type { TestSetupResult } from "../helpers/types";
 
 export const test = base.extend<{ testData: TestSetupResult }>({
-  // Default mode with large file: 1 x 200MB, uses existing case
+  // Default mode with large file: 1 x 50MB, uses existing case
   testData: [async ({ page }, use, testInfo) => {
     const result = await setupDefaultTestData(page, {
-      fileSizeMb: 200,
+      fileSizeMb: 50,
```
HasanCPS: ###  Make the large default-mode file size configurable

The normal default-mode fixture gets its size from `TEST_FILE_SIZE_MB`, but the large-file fixture still hardcodes `50`:

```ts
const result = await setupDefaultTestData(page, {
  fileSizeMb: 50,
  fileCount: 1,
});
```

That keeps CI stable today, but changing the large test to 25MB, 50MB, or 100MB still needs a code change. Since the rest of the suite already supports env-driven file sizes, it would be cleaner to add a separate large-file setting with the same default:

```ts
const result = await setupDefaultTestData(page, {
  fileSizeMb: config.largeTestFileSizeMb,
  fileCount: 1,
});
```

with something like:

```text
LARGE_TEST_FILE_SIZE_MB=50
```

That keeps CI and local runs aligned, while making runtime tuning a config change.

devops-pipelines/templates/e2e-playwright-test-job.yml
lilachdavis: changed to:
`- script: node_modules/.bin/playwright install --with-deps chromium`

devops-pipelines/templates/e2e-playwright-test-job.yml
lilachdavis: Added:
` env:
    CI: "true"`

e2e/pw/fixtures/test-fixtures-default-large.ts
lilachdavis: Done

e2e/pw/README.md
HasanCPS: ### Finish updating docs and labels for the configurable large-file size

The latest commit adds `LARGE_TEST_FILE_SIZE_MB` and uses it in the large default-mode fixture:

```ts
fileSizeMb: config.largeTestFileSizeMb,
```

Location: `e2e/pw/fixtures/test-fixtures-default-large.ts:13`

The README has not fully caught up:

```text
| `TEST_FILE_SIZE_MB` | Test file size in MB | No (default: 100) |
```

The actual default is now `10`, and the new `LARGE_TEST_FILE_SIZE_MB` setting is missing from the environment-variable table.

The test title is also still hardcoded to 50MB:

```ts
test.describe("Egress to NetApp Copy - Large File 50MB (Default Mode)", () => {
  test("should copy a 50MB file from Egress to NetApp using existing case", async ({
```

Location: `e2e/pw/tests/egress-to-netapp-copy-large-default.spec.ts:8`

If someone overrides `LARGE_TEST_FILE_SIZE_MB`, the test still reports as "50MB". That does not break the transfer, but it makes local and CI results misleading. I would update the README defaults and either keep the title generic, or include the configured size in a log line rather than the test name.

lilachdavis: Changes made accordingly

HasanCPS (approved):
LGTM!

PR #443, opened 2026-05-19
upgrade to .net 10

bridgesr:


lilachdavis (dismissed):
LGTM

devops-pipelines/backend/backend-build-and-deploy.yml
```diff
@@ -13,10 +13,10 @@ variables:
   - group: lacc-deployment-global
   - name: buildConfiguration
     value: 'Release'
-  - name: dotNetVersion
-    value: '8.x'
-  - name: efVersion
-    value: '8.0.*'
+- name: dotNetVersion
```
HasanCPS: ### Fix the broken variables block in the main backend deployment pipeline

Can we fix the indentation for the new `.NET 10` variables? `dotNetVersion` and `efVersion` are currently at the root level instead of under `variables`, and `asposeLicenceFileName` is then indented under the wrong block. That makes the YAML invalid, so the main backend build-and-deploy pipeline will fail to load after merge.

Current file:

```yaml
variables:
  - group: lacc-deployment-global
  - name: buildConfiguration
    value: 'Release'
- name: dotNetVersion
  value: '10.x'
- name: efVersion
  value: '10.0.*'
  - name: asposeLicenceFileName
    value: 'Aspose.Total.NET.lic'
```

I also checked it with a YAML parser:

```text
ERR devops-pipelines/backend/backend-build-and-deploy.yml:
while parsing a block mapping
expected <block end>, but found '-'
line 16, column 1
```

It should stay as one list under `variables`:

```yaml
variables:
  - group: lacc-deployment-global
  - name: buildConfiguration
    value: 'Release'
  - name: dotNetVersion
    value: '10.x'
  - name: efVersion
    value: '10.0.*'
  - name: asposeLicenceFileName
    value: 'Aspose.Total.NET.lic'
```

backend/CPS.ComplexCases.Common/Extensions/ResiliencePipelineExtensions.cs
```diff
@@ -0,0 +1,124 @@
...
+  public static ResiliencePipelineBuilder<HttpResponseMessage> AddStandardHttpResilience(
+      this ResiliencePipelineBuilder<HttpResponseMessage> pipeline,
+      HttpResilienceOptions options,
+      ILogger logger)
+  {
+    // Concurrency limiter outermost: shed excess load before doing any work.
+    if (options.ConcurrencyLimit > 0)
+    {
+      pipeline.AddConcurrencyLimiter(permitLimit: options.ConcurrencyLimit, queueLimit: int.MaxValue);
+    }
```
HasanCPS: ###  Concurrency limiter is now outermost, so a permit is held across the whole retry+backoff window

The concurrency limiter is now the outermost strategy, so a request holds its permit for the full retry sequence, backoff sleeps included. The old v7 wrap (`Policy.WrapAsync(retry, breaker, bulkhead)`) had the bulkhead innermost, so a slot was only held during a single attempt and released while the request waited to retry.

The comment says the limiter sits outermost to "shed excess load before doing any work", but `queueLimit: int.MaxValue` with no queue timeout means the limiter itself never rejects anything. It just queues.

Put those together and a burst of transient 5xx can stall throughput. With `ConcurrencyLimit = 30` on Egress, up to 30 requests sit in backoff holding every permit, and new requests queue behind them until their own per-operation timeout fires (`BaseEgressClient` wraps each call in a `CancellationToken`: 100s for management calls, the longer `TransferTimeoutSeconds` for transfers) and then fail. So under a sustained burst callers time out even though the downstream has spare capacity. Under the old innermost bulkhead the permit was released during each backoff wait, so a slow-retrying request didn't tie up a slot the whole time.

Was the outermost placement deliberate? If the intent is real load-shedding, a finite `queueLimit` (and probably a queue timeout) would shed instead of pile up. If the intent was to keep the old bulkhead behaviour, the limiter needs to sit inside the retry so the permit is freed during backoff.

backend/CPS.ComplexCases.NetApp/Factories/NetAppCertFactory.cs
```diff
@@ -28,7 +28,7 @@ public X509Certificate2Collection GetTrustedCaCertificates()
             try
             {
                 var rootCaBytes = Convert.FromBase64String(rootCaBase64);
-                var rootCaCert = new X509Certificate2(rootCaBytes);
+                var rootCaCert = X509CertificateLoader.LoadCertificate(rootCaBytes);
```
HasanCPS, line 31: This is a documented .NET breaking change, not just a rename. `X509CertificateLoader.LoadCertificate(bytes) ` only loads a single DER-encoded certificate and is intentionally stricter than the legacy `new X509Certificate2(bytes)`, which detected the content type and accepted additional formats such as PKCS#7 bundles. As a result, if any of the three Key Vault secrets (`RootCaCert, IssuingCaCert, IssuingCaCert2`) are stored in a format other than a raw DER certificate, the new implementation will throw a `CryptographicException` where the previous code would have succeeded. The exception is then rethrown as an `InvalidOperationException`, causing NetApp S3 startup to fail outside of Development.

Base64-encoded DER certificates, which is the expected format, are unaffected. The risk only exists if a secret is stored using a different encoding or container format. Before this change is released, it would be worth confirming what each environment's secrets actually contain. If any are stored as PEM or PKCS#7 bundles, they will require a different loading approach (`X509Certificate2.CreateFromPem` for PEM, or certificate collection import for PKCS#7).

It would also be worth reaching out to DevOps to get access to the testing environment credentials and validate this locally against the certificates currently in use. That would confirm the secret formats and help reduce the risk of any startup issues appearing after deployment.

backend/CPS.ComplexCases.API/Extensions/IServiceCollectionExtension.cs
```diff
@@ -25,29 +25,23 @@ public static void AddFileTransferClient(this IServiceCollection services, IConf
...
+
+            pipeline.AddRetry(new HttpRetryStrategyOptions
+            {
+                MaxRetryAttempts = retryAttempts,
+                Delay = TimeSpan.FromSeconds(firstRetryDelaySeconds),
+                BackoffType = DelayBackoffType.Exponential,
+                UseJitter = true,
+                ShouldHandle = static args =>
+                    ValueTask.FromResult(args.Outcome.Result?.StatusCode >= HttpStatusCode.InternalServerError)
+            });
```
HasanCPS: ### FileTransfer retry duplicates the shared helper and already retries POST/PUT on 5xx

This hand-rolls the same exponential-backoff-with-jitter retry that the new shared `AddStandardHttpResilience` already builds, and the two have drifted in a way that matters. The shared helper deliberately keeps POST and PUT out of status-code retries (`ExcludesPostAndPut`); this copy doesn't. So the FileTransfer client will retry a POST or PUT on a 5xx and can re-send a non-idempotent request, which risks duplicating whatever that call did on the transfer side.

One other difference: this only inspects `Result?.StatusCode`, so a connection failure (null result) isn't retried here, whereas the shared helper can retry those when `RetryOnConnectionFailure` is set.

Can this go through the shared helper instead of copying the retry? That keeps the POST/PUT exclusion and the connection-failure handling in one place. The blocker today is that the shared helper always forces a circuit breaker, so the two are worth sorting together.

backend/CPS.ComplexCases.Common/Extensions/ResiliencePipelineExtensions.cs
```diff
@@ -0,0 +1,124 @@
...
+              || options.AdditionalRetryableStatusCodes.Contains(response.StatusCode);
+
+          return ValueTask.FromResult(isRetryableStatus && ExcludesPostAndPut(response));
+        }
+      });
+    }
+
+    // Only "service is down" signals (5xx and connection failures) trip the breaker. Non-health
+    // signals such as 404 or 429 are deliberately excluded even when retry handles them.
+    pipeline.AddCircuitBreaker(new HttpCircuitBreakerStrategyOptions
```
HasanCPS: ### Shared helper always adds a circuit breaker with no opt-out

Retry is gated on `RetryAttempts > 0` and the limiter on `ConcurrencyLimit > 0`, but the circuit breaker is added unconditionally. There's no way to build a retry-only or limiter-only pipeline through this helper, which is the reason `AddFileTransferClient` can't reuse it and rolls its own retry ([see here](https://github.com/CPS-Innovation/Large-and-Complex-Cases/pull/443#discussion_r3603843280)).

If the breaker were gated like the others (a `CircuitBreakerEnabled` flag, or treating `MinimumThroughput <= 0` as off), the FileTransfer client could drop its duplicate and go through the shared path.

HasanCPS (approved):
LGTM!

PR #451, opened 2026-05-22
Feature/fct2 17261 polling

bridgesr:


backend/CPS.ComplexCases.API/Functions/Transfer/GetTransferStatus.cs
```diff
@@ -33,8 +33,12 @@ public async Task<IActionResult> Run(
         [HttpTrigger(AuthorizationLevel.Anonymous, "get", Route = "v1/filetransfer/{transferId}/status")] HttpRequest req, FunctionContext functionContext, string transferId)
     {
         var context = functionContext.GetRequestContext();
+        var ifNoneMatch = req.Headers["If-None-Match"].FirstOrDefault();
 
-        var response = await _transferClient.GetFileTransferStatusAsync(transferId, context.CorrelationId);
+        var response = await _transferClient.GetFileTransferStatusAsync(transferId, context.CorrelationId, ifNoneMatch);
+
+        if (response.Headers.TryGetValues("ETag", out var etags))
```
HasanCPS, line 40: I think we also need to expose the `ETag` header for browser clients.

Returning `ETag` from the API is only half of the flow. When the SPA calls the API from a different origin, the browser will not let frontend code read that header unless it is explicitly exposed through CORS.

So this can still return an `ETag`:

```http
ETag: abc123
```

but the UI can still see:

```javascript
response.headers.get("ETag") // null
```

If that happens, the UI never stores `lastEtag`, so it never sends `If-None-Match` on the next poll. The polling then keeps downloading full `200` responses instead of getting `304 Not Modified`.

Can we add this header on the API/gateway response as well?

```http
Access-Control-Expose-Headers: ETag
```

backend/CPS.ComplexCases.FileTransfer.API/Dtos/TransferStatusDto.cs
```diff
@@ -0,0 +1,41 @@
+using CPS.ComplexCases.Common.Models.Domain.Enums;
+using CPS.ComplexCases.FileTransfer.API.Durable.Payloads.Domain;
+using CPS.ComplexCases.FileTransfer.API.Models.Domain.Enums;
+
+namespace CPS.ComplexCases.FileTransfer.API.Dtos;
+
+public class TransferStatusDto
```
HasanCPS, line 7: Can we keep the existing non-sensitive status fields in this DTO, or update the Postman E2E collection in the same change?

The current Postman transfer-status polling still expects fields from the previous response contract, including `id` and `failedFiles`. This DTO now returns a smaller shape, so completed E2E journeys can fail even when the transfer itself succeeds.

The Postman location that also needs changing is:
`e2e/postman/LCCUserJourneyTests.postman_collection.json:2329`

That request logs `response.id`, logs `response.successfulFiles` / `response.failedFiles`, and asserts `response.failedFiles === 0`.

Suggested fix: either preserve the non-sensitive fields needed by the existing contract, such as `id`, `successfulFiles`, `failedFiles`, and timestamps, or update the Postman assertions at the location above to match this new DTO.

ui-spa/src/schemas/responses/transferStatusResponse.ts
```diff
@@ -11,6 +11,7 @@ export const transferFailedItemSchema = z.object({
 });
 
 export const transferStatusResponseSchema = z.object({
+  id: z.string().uuid().optional(),
```
HasanCPS: Can we keep these restored status fields required in the UI schema and update the Playwright mocks instead?

The backend DTO now returns `id`, `startedAt`, `successfulFiles`, and `failedFiles`, so the API contract has been restored. The latest change makes those fields optional in the UI schema, which appears to be because some inline Playwright status mocks do not include them.

Examples of mocks that need updating:
- `ui-spa/playwright/integration-tests/egress-netapp-transfer.spec.ts:433`
- `ui-spa/playwright/integration-tests/netapp-egress-transfer.spec.ts:430`
- `ui-spa/playwright/integration-tests/transfer-error-page.spec.ts:64`

Suggested fix: add the restored fields to those mocked status responses and keep the schema strict. Otherwise, the UI tests can pass even if the backend drops these fields again, while the Postman E2E contract would fail.

HasanCPS (approved):
LGTM!

PR #452, opened 2026-05-22
FCT2-17944 FCT2-17945 FCT2-17946 : Adressing the Gaps in postmann collection

kmshrajcps:
close the silent-pass bugs that let backend regressions slip through.

 PartiallyCompleted no longer passes silently.

Run-E2E-Tests.ps1:
- Change egressDestinationFolder from "4. Served Evidence/" (the upload folder, which guaranteed FileExists collision on every copy-back) to "3. Unused - disclosed/", aligning with the template default.

Setup-EgressWorkspaceAndUpload.ps1:
- Bump small-file post-upload verification window from 5x2s (10s total) to 20x3s (60s) so we more often capture Egress's canonical file ID instead of falling back to the upload-session ID, which trips up some downstream storage reads.

Verified: both `Run-E2E-Tests.ps1 -SizeMB 10 -RegisterCase` and the default-mode equivalent pass all three E2E journeys end-to-end.
Capture101
Capture100


e2e/postman/LCCUserJourneyTests.postman_collection.json
```diff
@@ -1508,215 +1160,280 @@
...
+									"    pm.environment.set('egressFileNames', sortedFiles.map(f => f.name).join(','));",
+									"    pm.environment.set('egressFileId', sortedFiles[0].id);",
+									"    pm.environment.set('egressFileName', sortedFiles[0].name);",
+									"    pm.environment.set('egressFileCount', sortedFiles.length.toString());",
+									"} else {",
+									"    console.log('[ERROR] No files found in folder!');",
+									"}",
+									"",
+									"// Clean up",
+									"pm.environment.unset('egressServedEvidenceFolderId');",
```
HasanCPS: This cleanup breaks the full Postman Runner flow.

`0. E2E Setup` now populates `egressServedEvidenceFolderId` once, then the Copy journey clears it here:

```javascript
pm.environment.unset('egressServedEvidenceFolderId');
```

The Move journey still uses the same variable in its first list request:

```text
folder-id={{egressServedEvidenceFolderId}}
```

So when the whole collection runs end to end, Move can start with an empty folder id and fail to find its source file.

Can we keep `egressServedEvidenceFolderId` until the Move list request has used it, or repopulate it per journey before clearing it?

e2e/postman/LCCUserJourneyTests.postman_collection.json
```diff
@@ -7560,35 +2957,42 @@
...
 									"    console.log('Valid files (non-zero size):', validFiles.length);",
 									"    ",
-									"    if (validFiles.length > 0) {",
-									"        // Pick the most recent file",
-									"        const selectedFile = validFiles[0];",
+									"    // Deterministically pick the file Egress->NetApp Copy deposited this run.",
+									"    // 'latest by lastModified' was unsafe -- it picked stale files from prior",
+									"    // runs whose namesakes already exist in Egress 3. Unused - disclosed/,",
+									"    // producing PartiallyCompleted at destination.",
+									"    const expectedName = pm.environment.get('egressCopyFileName');",
```
HasanCPS, line 3106: This makes the standalone `netapp-to-egress` run depend on Copy journey state.

When we run:

```powershell
.\Run-E2E-Tests.ps1 -TestsToRun netapp-to-egress
```

the script does not upload a Copy source and does not run the Egress-to-NetApp Copy folder, so `egressCopyFileName` is not set. The selection logic then fails before it can choose an existing NetApp file.

Can we keep a safe fallback for standalone `netapp-to-egress`, or make sure this value is created by the setup path for that mode?

e2e/postman/Run-E2E-Tests.ps1
```diff
@@ -801,105 +808,82 @@ Write-Host ""
...
+    foreach ($folder in $foldersToRun) {
+        Write-Host ""
+        Write-Host "─────────────────────────────────────────────────────────────" -ForegroundColor DarkGray
+        Write-Host "Preparing: $folder" -ForegroundColor Cyan
+        Write-Host "─────────────────────────────────────────────────────────────" -ForegroundColor DarkGray
+
+        $result = Run-NewmanFolder `
+            -CollectionPath $UpdatedCollectionPath `
+            -FolderName $folder `
+            -EnvironmentPath $postSetupEnvPath `
```
HasanCPS, line 871: I think we need to refresh auth before the later Newman journeys.

With the new shared setup, each journey reuses the same `postSetupEnvPath` created before the first transfer. That is fine for shorter runs, but for a long `-TestsToRun all` run the later folders can start with an old `lccAccessToken`.

Example:

```powershell
.\Run-E2E-Tests.ps1 -TestsToRun all -SizeGB 1 -FileCount 3
```

If Copy takes a while, NetApp-to-Egress or Move may start after the original bearer token has expired. The collection-level pre-request script does refresh `accessToken` and CMS auth, but the journey requests use `Authorization: Bearer {{lccAccessToken}}`, and I can only see `lccAccessToken` being set in the setup folder.

Can we refresh `lccAccessToken` before each Newman journey, or update the shared auth refresh so the token used by the journey request headers is refreshed as well?

e2e/postman/Setup-EgressWorkspaceAndUpload.ps1
HasanCPS, line 712: Can we avoid relying on the upload ID fallback here and make sure we resolve the canonical Egress file ID before continuing?

In local testing the upload completes, but the verification polls 20 times and then falls back to `$UploadId`. That looks risky because the upload uses `folder_path`, while this check lists by `folder_id=$FolderId`, and `$FolderId` may be empty. The fallback can then pass an upload-session ID into later flows that expect a real file ID, which can cause `404` errors on move.

Could we either use the file ID from the completion response if Egress returns it, or resolve the target folder ID before polling?

Image

e2e/postman/LCCUserJourneyTests.postman_collection.json
```diff
@@ -1,8587 +1,3966 @@
...
-                "key": "Authorization",
-                "value": "Bearer {{lccAccessToken}}"
-              },
-              {
-                "key": "Cookie",
-                "value": "Cms-Auth-Values={{cmsAuth}}"
-              },
-              {
-                "key": "Content-Type",
-
```
HasanCPS, line 1643: Can we make the NetApp copy destination unique per collection run?

This request still writes the Egress-to-NetApp copy to `{{netappFolderPath}}`. If the imported collection is run again with the same generated environment, the same file name already exists in NetApp and the backend can return `PartiallyCompleted` with `FileExists`.

Example fix: use a run-scoped path such as `Automation-Testing/e2e-<run-id>/` and use the same value for validate and initiate.

HasanCPS, line 2196: Can we fail the Move journey when the expected move source is missing instead of falling back to the latest files?

On a no-upload rerun, the original Move source has already been moved out of `4. Served Evidence/`. This fallback can then pick the remaining Copy source and move/delete the wrong file while still reporting success.

Expected behavior: only use files matching the configured Move source, for example `egressMoveFileNames`. If they are not present, the test should fail or re-upload a Move source.

HasanCPS, line 3337: Can we make the NetApp-to-Egress destination unique per collection run?

This copy-back request still writes to the fixed `{{egressDestinationFolder}}`. After the first run, the file already exists under `3. Unused - disclosed/`, so a second run with the same environment can return `PartiallyCompleted` with `FileExists`.

Example fix: set `egressDestinationFolder` to a per-run subfolder such as `3. Unused - disclosed/e2e-<run-id>/` before validate, then reuse that value for initiate.

HasanCPS, line 3106: Can we add copy-back coverage for the Move journey as well?

The current NetApp-to-Egress copy-back flow selects `egressCopyFileName` from `Automation-Testing/`. The Move journey writes to `Automation-Testing-Move/`, so the moved file is not copied back to Egress or asserted.

Suggested fix: add a separate Move copy-back folder, or parameterise this flow so it can copy back and verify the Move output too.

e2e/postman/LCCUserJourneyTests.postman_collection.json
```diff
@@ -1,8587 +1,4865 @@
...
```
HasanCPS: This `pm.execution.setNextRequest(null)` ends the whole collection run, so in a full Postman Runner pass the next folder, `E2E: Netapp Move to Egress Copy` (line 3802), never executes. That folder is the Move round-trip coverage the ticket asks for, so from "import and hit Run" the new journey is silently skipped. Newman is unaffected because each folder is a separate invocation.

Can we remove the `setNextRequest(null)` here and let the run fall through (the ENC/ENM journey-end scripts already do this correctly), keeping it only on the genuinely-last `[NME] 14. Download Activity Log`?

While here: `14.`/`[NME] 14.` don't assert the final transfer status, but `[ENC] 12.`/`[ENM] 12.` do (`Final transfer status is Completed`). Can we add the same assertion so all four journeys end with the same strictness?

HasanCPS: The Copy-vs-Move file-set selection works by regex-matching the folder name out of `pm.execution.location`. If that API is unavailable in the Newman version running in CI, or the folder is ever renamed, `_inMove` silently evaluates to `false` and the Move journey selects the `egressCopy*` file set — meaning **Move would move/delete the Copy source**, which is the exact regression the per-journey file split exists to prevent.

The ENC and ENM `7b` scripts are already two separate copies (they differ only in the final `setNextRequest`), so the journey prefix can simply be hardcoded in each folder's copy. Can we replace the location-sniffing with a hardcoded `'egressCopy'` / `'egressMove'` per folder?

HasanCPS: The request-level guard skips only when `registerCase === 'false'`, so a manual Send of this request with the variable missing/blank (and no `caseId` set) still registers a case in CMS — one of the entry points the ticket explicitly lists ("clicking Send on any Register Case request individually"). The folder-level guard in `1. Get Azure AD Token` already uses the positive `!== 'true'` check; can we make this request-level guard match it?

```js
if (_registerCase !== 'true') { ... skip ... }
```

HasanCPS: The uniquify suffix is `ws-{{egressWorkspaceId}}`, which is unique per **workspace**, not per **run** — but the comments claim it prevents FileExists on "a second import-and-run cycle against the same generated env". A replayed env has the same workspace id and the same `egressCopyFileName`, so the second run writes the same file to the same path. The backend then fails it: `TransferFile.cs:63-77` pre-checks the NetApp destination and raises `FileExists`, and `TransferState.cs:30` maps any failed item to `PartiallyCompleted` — which the (correctly) strengthened poll assertion now fails. So the ticket's "Runner twice in a row must pass" criterion is not met; in default mode (shared `DefaultWorkspaceId`) the suffix adds nothing run-to-run.

Two findings from checking the backend:

1. A true per-run subfolder is feasible with no backend change: `EgressStorageClient` auto-creates missing destination folder structure on upload (`EgressStorageClient.cs:76-80`), and NetApp destinations are S3-style keys where prefixes are implicit. So `0. E2E Setup` could stamp a run id once (e.g. `pm.environment.set('e2eRunId', pm.variables.replaceIn('{{$guid}}'))`, exported in the post-setup env) and the suffix becomes `e2e-run-<id>/` — genuinely unique per run, consistent across all journey folders.
2. The Move journey can never satisfy "second Runner run passes without re-running PowerShell" regardless of destination uniqueness, because the backend deletes the Egress source on Move (`TransferOrchestrator.cs:243-248`). Run 2's `[ENM] 7b` will correctly fail loudly on the missing source. Unless we add an in-collection re-upload, I suggest we agree that this AC applies to the Copy journeys only and document the Move limitation in the README.

At minimum, can we correct the comment blocks so they don't claim replay-safety the suffix doesn't provide?

One more nuance: for the NetApp→Egress direction the `FileExists` pre-check does **not** run (`TransferFile.cs:63` guards it to `EgressToNetApp` only), so whether a duplicate in `3. Unused - disclosed/` actually fails depends on the external Egress API, not our code. Worth verifying empirically before relying on it either way.

HasanCPS: This logging block reads `response.fileStatuses`, but that field doesn't exist in the transfer status contract — the branch returns `TransferEntity` and main returns `TransferStatusDto`, and both expose failures as `failedItems[{sourcePath, errorCode}]`. The ENC/ENM poll scripts already log `failedItems` correctly. Can we switch this block to `failedItems` so partial failures in the NetApp→Egress journeys actually log the failing files instead of silently logging nothing?

HasanCPS: `transferPollCount` is incremented but never checked against a maximum, for both the 404-retry path and the `InProgress` path. A transfer that never reaches a terminal state (or a transferId that 404s forever) loops indefinitely at ~3s per iteration until the CI job timeout kills the run with no useful diagnostics. Since the poll scripts are being touched anyway, can we add a cap (e.g. fail the test after N attempts with the last seen status)? Applies to all four journeys' poll scripts.

HasanCPS: In standalone `-TestsToRun netapp-move-to-egress` (or a Runner pass where the Move journey didn't run), `egressMoveFileName` is set but the file was never deposited in `Automation-Testing-Move/ws-<id>/`, so this constructed-path fallback submits a transfer for a file that doesn't exist and the journey fails later with a confusing transfer error. The fallback's "listing hasn't caught up yet" assumption only holds when the Move journey ran in the same top-level run. Can we either fail fast here with a clear "Move journey has not deposited the source — run `move` first" message, or document that `netapp-move-to-egress` requires `move` to have run against the same workspace?

e2e/postman/LCCUserJourneyTests.postman_collection.json
```diff
@@ -1,8587 +1,4969 @@
...
```
HasanCPS, line 3991: The fail-fast here marks the test failed but doesn't stop the journey. Nothing calls `pm.execution.setNextRequest(null)` after the failing assertion, and `Run-E2E-Tests.ps1` runs Newman without `--bail`, so `[NME] 10` through `[NME] 14` still execute. In a full Runner pass where the Move forward folder was filtered out, `netappFilePath` still holds the Copy journey's file (set by `9. List Netapp Files` in the previous folder), so this journey submits a transfer for the wrong file. That's the cross-journey mix-up the per-journey split was built to prevent. It's a copy rather than a move, so nothing is deleted, and the run still ends up red, but we've fired an unwanted transfer and buried the real failure under downstream noise.

Can we add `pm.execution.setNextRequest(null)` straight after the failing test so the journey actually stops here?

HasanCPS: `MAX_POLL_ATTEMPTS = 60` works out at roughly 3 to 4 minutes per transfer: each poll costs the 2000ms busy-wait in the prerequest, plus Newman's `--delay-request 1000`, plus the request itself. The suite explicitly supports `-SizeGB` (default 1 GB) and `-FileCount`, and a transfer that size can legitimately take longer than 4 minutes, so runs that passed before this change could now fail at the cap rather than on anything real.

Could we read the cap from the environment with 60 as the default, e.g. `parseInt(pm.environment.get('maxPollAttempts') || '60')`, and let `Run-E2E-Tests.ps1` scale it with the requested file size?

HasanCPS: Same constant at Line numbers `2635, 3500, 4348`

HasanCPS, line 1701: When the cap trips, neither the 404 branch nor the InProgress branch calls `setNextRequest`, so execution falls through to `Verify Activity Log` and `Download Activity Log`, which then fail as well ("Final transfer status is Completed"). One root cause shows up as three failures in the report. Worth ending the journey at the cap so the report points at the actual problem?

HasanCPS: Same pattern in the other three poll scripts

e2e/postman/Run-E2E-Tests.ps1
```diff
@@ -723,7 +724,14 @@ $variables = @{
     "registerCaseClientId" = $Config.RegisterCaseClientId
     "lccApiId" = $Config.LccApiId
     "netappFolderPath" = "Automation-Testing/"
-    "egressDestinationFolder" = "4. Served Evidence/"
+    # NetApp -> Egress copy-back destination. Must NOT equal the upload
+    # folder (`4. Served Evidence/`) or every run produces PartiallyCompleted
+    # with FileExists, because the file being copied back from NetApp is the
+    # one the upload step just put in `4. Served Evidence/`. Per issue #4,
```
HasanCPS: This comment block still explains FileExists-avoidance through timestamped upload filenames ("Per issue #4 ... collisions there only occur if the same timestamped filename is reused"). The collection now appends `e2e-run-<id>/` to `egressDestinationFolder`, and that subfolder is the actual guard. Can we update this block to describe the current mechanism?

e2e/postman/LCCUserJourneyTests.postman_collection.json
```diff
@@ -1,8587 +1,5004 @@
...
```
HasanCPS: > The `[NME] 9` fix in 586ae48 didn't get mirrored here. This script's empty-folder
> branch (line 3151) marks the test failed but doesn't end the journey, and the
> trailing unconditional `setNextRequest('10. Validate Transfer NetApp to Egress')`
> would override a `null` even if the branch set one. With the folder empty and a
> stale `netappFilePath` in a replayed env, requests 10–14 submit a transfer for
> whatever file the previous run used. Lower stakes than the Move journey (nothing
> gets deleted, and the full E2E flow always deposits a file here first), but for
> consistency: drop the trailing `setNextRequest` and add
> `pm.execution.setNextRequest(null)` to the empty-folder branch, same as `[NME] 9`.

HasanCPS, line 3091: > The new fail-fast branches only run inside `if (pm.response.code === 200)`. On a
> 401/500 from the list endpoint, the 'Status code is 200' test fails but nothing
> ends the journey, so 10–14 run with whatever `netappFilePath` is left in the env —
> the same stale-file transfer the 586ae48 fix prevents for the in-200 branches. An
> `else { pm.execution.setNextRequest(null); }` on the status check would close the
> gap.

HasanCPS, line 3949: > The new fail-fast branches only run inside `if (pm.response.code === 200)`. On a
> 401/500 from the list endpoint, the 'Status code is 200' test fails but nothing
> ends the journey, so 10–14 run with whatever `netappFilePath` is left in the env —
> the same stale-file transfer the 586ae48 fix prevents for the in-200 branches. An
> `else { pm.execution.setNextRequest(null); }` on the status check would close the
> gap.

e2e/postman/Setup-EgressWorkspaceAndUpload.ps1
HasanCPS: **Upload verification can never find the file, so it always falls back to the upload-session ID**

The verify loop lists workspace files filtered by `folder_id`:

```powershell
.../workspaces/$WorkspaceId/files?folder_id=$FolderId
```

`$FolderId` is set to `""` at lines 410 and 419 and never assigned anywhere after that. Uploads target the destination with `folder_path` (line 464), not `folder_id`, so this listing never returns the file we just uploaded. Every poll logs `not indexed yet`, all 20 attempts miss, and the script drops to the upload-session ID at line 718.

On a clean 10MB run both files go 0-for-20 and end up with the upload-session ID instead of the canonical Egress file ID. Copy tolerates that ID, but Move fails downstream with a 404 from `EgressStorageClient.OpenReadStreamAsync`, which can't resolve an upload-session ID to a blob. It also burns the whole 60s polling window per file, so the "1.1 min @ 0.16 MB/s" in the summary is the dead loop, not the actual upload.

Fix is to query by `folder_path`, matching how the file was uploaded and how the collection's own List Egress Files request already does it (`?folder_path={{egressUploadFolderPath}}`, line 1121). The path has spaces, so encode it:

```powershell
$encodedFolder = [uri]::EscapeDataString($ActualFolderPath)
$filesResult = curl.exe --silent --location "$BaseUrl/api/v1/workspaces/$WorkspaceId/files?folder_path=$encodedFolder" `
    --header "Authorization: Basic $TokenBase64"
```

Minor, while you're in here: the loop sleeps before its first check (line 689), so even a file that indexes instantly costs the first delay. Swapping to check-first, sleep-on-miss shaves that off.

e2e/postman/Setup-EgressWorkspaceAndUpload.ps1
```diff
@@ -673,15 +673,28 @@ if ($SkipUpload) {
...
+        # flows (the Copy journey) but trips up others (the Move journey hits
+        # 404 from EgressStorageClient.OpenReadStreamAsync) because Egress's
+        # storage layer can't always resolve an upload-session ID into a blob.
+        # 10s (the previous small-file default of 5x2s) was reliably too short.
+        $retryCount = if ($IsLargeFile) { $MaxFileIdRetries } else { 20 }
+        $retryDelay = if ($IsLargeFile) { $FileIdRetryDelaySeconds } else { 3 }
+
+        # Verify by folder_path -- uploads above use folder_path, not folder_id,
+        # so the listing has to match. Path is URL-encoded (spaces).
+        $encodedFolder = [uri]::EscapeDataString($ActualFolderPath)
```
HasanCPS: ## Verify loop never finds the file, so it falls back to the upload ID (breaks the Move journey)

On a clean run the upload completes, but the verify loop polls all 20 times, logs `not indexed yet` every attempt, and drops to `$UploadId`. That upload-session ID isn't a real file ID, so the Copy journey limps along but the Move journey 404s in `EgressStorageClient.OpenReadStreamAsync`, which can't resolve an upload-session ID to a blob.

Two things are wrong in the list call:

1. **Query param.** It lists with `?folder_path=$encodedFolder`. Egress has no `folder_path` (or `folder_id`) filter for this. The working contract is `?path=<folder>` — see `Clear-EgressFolder.ps1:47`, which is the same Egress list endpoint and runs in the cleanup pipeline. The folder must have no trailing slash, and spaces have to be encoded while `/` stays literal. `[uri]::EscapeDataString` on the whole path turns `/` into `%2F` and breaks the match, so encode per segment.

2. **Response parse.** The code reads the top level as an array and uses `.name` / `.isFolder`. Egress wraps results in `{ "data": [...] }` and the entries look like this (captured from a live run):

   ```json
   { "_links": { "download": "...", ... }, "id": "6a32cb380b11b5e318526ec0", "filename": "copy-10MB-...-file1.txt" }
   ```

   So it's `$filesObj.data`, the name field is `filename`, and the canonical file ID is `id`. There's no `isFolder` and no `filesize` field here, so any check against those never matches.

Suggested fix for the verify block:

```powershell
$folderForQuery = $ActualFolderPath.TrimEnd('/')
$encodedFolder = ($folderForQuery -split '/' | ForEach-Object { [uri]::EscapeDataString($_) }) -join '/'

for ($i = 1; $i -le $retryCount; $i++) {
    if ($i -gt 1) { Start-Sleep -Seconds $retryDelay }
    Write-Host "  Attempt $i/$retryCount..." -NoNewline

    $filesResult = curl.exe --silent --location "$BaseUrl/api/v1/workspaces/$WorkspaceId/files?path=$encodedFolder" `
        --header "Authorization: Basic $TokenBase64"

    try {
        $filesObj = $filesResult | ConvertFrom-Json
        $uploadedFile = $filesObj.data | Where-Object { $_.filename -eq $CurrentFileName } | Select-Object -First 1

        if ($uploadedFile) {
            $FileId = $uploadedFile.id
            $ActualFilePath = if ($uploadedFile.path) { $uploadedFile.path + "/" } else { $ActualFolderPath }
            Write-Host " FOUND ($FileId)" -ForegroundColor Green
            break
        } else {
            Write-Host " not indexed yet" -ForegroundColor Gray
        }
    } catch {
        Write-Host " error" -ForegroundColor Yellow
    }
}
```

Note on matching: the listing has no `upload_id`, so we can't key off the upload ID here. Filename is the practical key, which is safe because the names are timestamped per upload and the folder is cleared per run. If we ever want a guaranteed upload-to-file mapping, the completion response (`PUT .../uploads/{id}/`) is the only place that link could live, and it's worth checking what that returns before relying on it.

HasanCPS: ## `e2e/postman/LCCUserJourneyTests.postman_collection.json:1121` — same broken param in the manual "List Egress Files" verify request

This request uses `?folder_path={{egressUploadFolderPath}}` against `{{egressBaseUrl}}`, which is the precedent the setup script copied. Same fix: use `?path={{egressUploadFolderPath}}` and update the query key from `folder_path` to `path`.

HasanCPS (approved):
LGTM!

PR #457, opened 2026-05-28
Task/fct2 18662 update cleanup to remove e2 e test files in shared drive

lilachdavis:
- Add script to remove files from a shared drive folder, with optional filtering by name, utilising the batch-delete function.
- Add helper module for shared drive cleanup.
- Update Clear-SharedDriveFolder to use helper module
- Update cleanup pipeline to clear the contents of existingCaseAutomation shared drive folder - where the E2E playwright test uploads files. The task is configured to always leave a single file in the folder, as per the tests' requirement.


devops-pipelines/cleanup-testfiles.yml
```diff
@@ -112,18 +113,18 @@ jobs:
...
+        displayName: 'Cleanup E2E tests NetApp folder'
+        condition: succeededOrFailed()
+        inputs:
+          filePath: '$(System.DefaultWorkingDirectory)/devops-pipelines/scripts/test-automation-cleanup/Remove-SharedDriveFolderContent.ps1'
+          arguments: >
+            -ClientId $(CallingAppValidAudience)
+            -TenantId $(TenantId)
+            -AadUsername $(AadUserName)
+            -AadPassword $(AadUserPassword)
+            -CaseId $(E2eTestsCaseId)
```
HasanCPS: I think this cleanup step is using the wrong case id variable.

The new NetApp cleanup passes:

```yaml
-CaseId "$(E2eTestsCaseId)"
```

but the active Playwright E2E setup uses the singular variable:

```yaml
DEFAULT_CASE_ID: "$(E2eTestCaseId)"
```

If only `E2eTestCaseId` is defined in the variable group, the cleanup script will get an empty or unresolved `-CaseId` value and fail before it deletes the NetApp test files.

Can we switch this cleanup step to use `$(E2eTestCaseId)`, or make sure both variable names are deliberately populated?

lilachdavis: Playwright E2E pipeline updated to use the same plural for as the cleanup pipeline $(E2eTestsCaseId) which is the correct key of the variable in the library.

HasanCPS (approved):
LGTM!

PR #459, opened 2026-06-01
Story/FCT2-18471  React Query

renjithabby:
https://cpsgovuk.atlassian.net/browse/FCT2-18471
- Left the transfer of materials un touched due to new design implementation in new sprint


ui-spa/src/components/egress-connect/index.tsx
```diff
@@ -28,11 +28,16 @@ const EgressPage = () => {
...
-    getEgressSearchResults,
-    [workspaceName],
-    false,
-  );
+  const { data: egressSearchResults, isLoading: isEgressSearchResultLoading } =
+    useQuery({
+      queryKey: [workspaceName],
+      queryFn: () => getEgressSearchResults(workspaceName),
+      retry: false,
+      enabled: true,
```
HasanCPS: I think we should only run this query once `workspaceName` has a value.

At the moment the first render uses the initial state of `""`, so the page can call:

```http
GET /api/v1/egress/workspaces?workspace-name=&skip=0&take=50
```

before the value from the URL has been applied. If Egress rejects that, `throwOnError` can send the user to the error boundary. If it returns broad results, the page may briefly show results for an empty search.

Can we change the query to only run when the workspace name is present, for example:

```ts
enabled: !!workspaceName
```

ui-spa/src/components/search-result-page/index.tsx
```diff
@@ -345,14 +344,8 @@ const CaseSearchResultPage = () => {
...
   };
-
-  if (
-    ((caseSearchApi.status === "loading" ||
-      caseSearchApi.status === "initial") &&
-      !errorList.length) ||
-    (formData[SearchFormField.searchType] !== "urn" &&
-      !formattedAreaValues.options.length)
-  ) {
+  console.log("searchResults", searchResults);
```
HasanCPS: Can we remove this console log, or keep it behind a local debug guard?

It logs the full `searchResults` payload whenever the search succeeds. That payload includes `urn`, `caseId`, `operationName`, `leadDefendantName`, `egressWorkspaceId` and `netappFolderPath`, so we should not write it to the browser console in normal builds.

HasanCPS (approved):
LGTM!

PR #464, opened 2026-06-10
FCT2-19239: Display Egress workspace in activity log when connecting/…

NeilFoubister:
…disconnecting.

### DIFF
--- a/backend/CPS.ComplexCases.API/Functions/CreateEgressWorkspace.cs
+++ b/backend/CPS.ComplexCases.API/Functions/CreateEgressWorkspace.cs
@@ -98,7 +98,7 @@ public class CreateEgressWorkspace(
         {
             CaseId = request.Value.CaseId,
             EgressWorkspaceId = workspace.Id,
-            EgressWorkspaceName = workspaceName
+            EgressWorkspaceName = workspace.Name
         });
 
         await _activityLogService.CreateActivityLogAsync(
--- a/backend/CPS.ComplexCases.API/Functions/CreateEgressWorkspace.cs
+++ b/backend/CPS.ComplexCases.API/Functions/CreateEgressWorkspace.cs
@@ -97,7 +97,8 @@ public class CreateEgressWorkspace(
         await _caseMetadataService.CreateEgressConnectionAsync(new CreateEgressConnectionDto
         {
             CaseId = request.Value.CaseId,
-            EgressWorkspaceId = workspace.Id
+            EgressWorkspaceId = workspace.Id,
+            EgressWorkspaceName = workspaceName
         });
 
         await _activityLogService.CreateActivityLogAsync(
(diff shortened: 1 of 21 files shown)


backend/CPS.ComplexCases.Data/Models/Requests/CreateEgressConnectionDto.cs
```diff
@@ -8,4 +8,6 @@ public class CreateEgressConnectionDto
   public required int CaseId { get; set; }
   [JsonPropertyName("egressWorkspaceId")]
   public required string EgressWorkspaceId { get; set; }
+  [JsonPropertyName("egressWorkspaceName")]
+  public required string EgressWorkspaceName { get; set; }
```
HasanCPS, line 12: > Making `EgressWorkspaceName` required (both the C# `required` keyword and the validator rule) turns this into a breaking change for the connect endpoint. Any request without `egressWorkspaceName` in the body now gets a 400.
>
> That's fine once the new UI is everywhere, but there are two windows where it bites:
>
> 1. If the API deploys before the UI.
> 2. Users who already have the SPA open in a tab. They keep the old bundle until they refresh, and their connect requests will fail with a 400 in the meantime.
>
> If our deploys aren't atomic, it might be safer to accept the field as optional for this release (fall back to the workspace ID for the log entry, which is what happens today anyway) and tighten it to required in a follow-up once the old bundle has aged out.

backend/CPS.ComplexCases.API/Functions/CreateEgressWorkspace.cs
```diff
@@ -97,15 +97,16 @@ public async Task<IActionResult> Run([HttpTrigger(AuthorizationLevel.Anonymous,
...
         });
 
         await _activityLogService.CreateActivityLogAsync(
           ActivityLog.Enums.ActionType.ConnectionToEgress,
           ActivityLog.Enums.ResourceType.StorageConnection,
           request.Value.CaseId,
           workspace.Id,
-          workspace.Id,
+          workspace.Name,
           context.Username);
```
HasanCPS, line 110: > Small inconsistency: the metadata DTO stores the locally generated `workspaceName`, but the activity log entry uses `workspace.Name` from the Egress create response. They should be the same string, but if Egress ever trims or normalises the name on creation, the database and the activity log would quietly disagree.
>
> Suggest using `workspace.Name` for both, since that's the name Egress actually created:
>
> ```csharp
> await _caseMetadataService.CreateEgressConnectionAsync(new CreateEgressConnectionDto
> {
>     CaseId = request.Value.CaseId,
>     EgressWorkspaceId = workspace.Id,
>     EgressWorkspaceName = workspace.Name
> });
>

backend/CPS.ComplexCases.Data/Models/Requests/CreateEgressConnectionDto.cs
NeilFoubister: @HasanCPS so remove the validator check too?

HasanCPS: fine to keep as it is.

HasanCPS (approved):
LGTM!

PR #469, opened 2026-06-15
FCT2-16843: Initial commit - ONTAP API integration for rename material.

NeilFoubister:


backend/CPS.ComplexCases.NetApp/Client/OntapHttpClient.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        var response = await CallOntap(request);
+
+        return response.StatusCode switch
+        {
+            HttpStatusCode.OK => new OkResult(),
+            HttpStatusCode.NotFound => new NotFoundObjectResult($"Material not found at path: {currentFolderPath}"),
+            HttpStatusCode.Conflict => new ConflictObjectResult($"Conflict occurred while renaming material from {currentFolderPath} to {newFolderPath}"),
+            _ => throw new OntapClientException(response.StatusCode, new HttpRequestException($"Unexpected status code: {response.StatusCode}")),
+        };
```
HasanCPS: ### The 404 and 409 branches never run in production

`RenameMaterialAsync` calls `CallOntap(request)` on line 35 with no expected status codes, then switches on the response and expects to handle `NotFound` (line 40) and `Conflict` (line 41). Those two branches are unreachable.

`CallOntap` forwards to the shared `HttpResponseHandler`. When you pass no expected status codes, the handler treats a 404 or 409 as a failure and throws, because both are registered in `StatusCodeExceptionFactories` for this client (`OntapNotFoundException`, `OntapConflictException`). So the response never comes back to the switch. The endpoint can only return 200 or, via the thrown exception, 500.

The unit tests pass because `OntapHttpClientTests` mocks `IHttpResponseHandler` and has the mock *return* a 404 or 409 response, which skips the throwing the real handler does. One test in that file, `RenameMaterialAsync_PassesExpectedStatusCodeExceptionFactories_ToResponseHandler`, even asserts that those throwing factories are registered, so the suite encodes two contradictory expectations and still goes green.

Pick one model and make it consistent:
- If you want the `NotFoundObjectResult` / `ConflictObjectResult` responses, pass the codes as expected: `CallOntap(request, HttpStatusCode.NotFound, HttpStatusCode.Conflict)`, and drop those two entries from `StatusCodeExceptionFactories`.
- If you want the exception model, delete the unreachable switch branches and let the middleware turn the exceptions into responses.

Either way, add a test that runs the real `HttpResponseHandler` for the 404 and 409 paths. The mock hides exactly the behaviour that breaks here.

backend/CPS.ComplexCases.API/Validators/Requests/MaterialRenameRequestValidator.cs
```diff
@@ -0,0 +1,21 @@
...
+        RuleFor(x => x.CaseId)
+            .GreaterThan(0).WithMessage("CaseId must be greater than 0.");
+
+        RuleFor(x => x.CurrentPath)
+            .NotEmpty().WithMessage("Current material path is required.")
+            .MaximumLength(260).WithMessage("Current material path cannot exceed 260 characters.");
+
+        RuleFor(x => x.NewPath)
+            .NotEmpty().WithMessage("New material path is required.")
+            .MaximumLength(260).WithMessage("New material path cannot exceed 260 characters.");
```
HasanCPS: ### Paths aren't scoped or sanity-checked

The validator covers non-empty and a 260 character cap, and nothing else. Two gaps:

- No `CurrentPath != NewPath` rule, so a no-op rename round-trips to ONTAP for nothing.
- No guard against `..` or a leading `/`. The paths go straight into the ONTAP file URL. Access is already scoped to the user's volume UUID, so the blast radius is small, but `DeleteNetAppBatchRequestValidator` already rejects `..` and leading slashes for the same kind of input, so it's worth matching that here.

The `Uri.EscapeDataString` encoding in `OntapRequestFactory.cs:12` is correct for ONTAP's `%2F`-separated file paths, so that part is fine.

backend/CPS.ComplexCases.NetApp/Exceptions/OntapConflictException.cs
```diff
@@ -0,0 +1,18 @@
...
+        : base("Conflict occurred while accessing ONTAP API.")
+    {
+    }
+
+    public OntapConflictException(string message)
+        : base(message)
+    {
+    }
+
+    public static HttpStatusCode StatusCode => HttpStatusCode.Conflict;
```
HasanCPS: ### `backend/CPS.ComplexCases.NetApp/Exceptions/OntapConflictException.cs:17` vs `OntapClientException.cs:7` — inconsistent `StatusCode` shape

`OntapConflictException` and `OntapUnauthorizedException` expose a `static` `StatusCode`, while `OntapClientException` exposes an instance one. The static version can't be read off a caught instance, so if the middleware ever switches on these it won't behave the way the static property suggests. Pick one shape across the set.

backend/CPS.ComplexCases.NetApp/Models/Dto/MaterialRenameDto.cs
```diff
@@ -0,0 +1,9 @@
+using System.Text.Json.Serialization;
+
+namespace CPS.ComplexCases.NetApp.Models.Dto;
+
+public class MaterialRenameDto
```
HasanCPS, line 5: ### Two classes named `MaterialRenameDto`

`NetApp/Models/Dto/MaterialRenameDto.cs` is the ONTAP payload (`{ path }`) and `NetApp/Models/Requests/MaterialRenameDto.cs` is the inbound request body. Both are reasonable, but the name collision is easy to trip over when reading. Consider renaming the inbound one to `RenameMaterialRequest`.

backend/CPS.ComplexCases.NetApp/Models/Requests/MaterialRenameDto.cs
```diff
@@ -0,0 +1,13 @@
...
+namespace CPS.ComplexCases.NetApp.Models.Requests;
+
+public class MaterialRenameDto
+{
+    [JsonPropertyName("caseId")]
+    public required int CaseId { get; set; }
+    [JsonPropertyName("currentPath")]
+    public required string CurrentPath { get; set; }
+    [JsonPropertyName("newPath")]
+    public required string NewPath { get; set; }
```
HasanCPS: ### This only renames one material per call

The request carries a single `CurrentPath` and `NewPath`, and that one-to-one shape runs the whole way down: the arg factory builds one `MaterialRenameArg`, the request factory produces one PATCH, and `RenameMaterialAsync` fires one call. A caller who wants to rename several materials has to make N separate requests.

If batch rename is in scope for this ticket or a near follow-up, this is the point to decide, because retrofitting it later means changing the DTO, the validator, the client method, and the function all at once. We already have the pattern for it in `DeleteNetAppBatch`, so the cleanest move is to mirror that rather than invent a new shape:

- DTO: `CaseId` plus `List<RenameNetAppBatchOperationDto>`, each item carrying `Type`, `CurrentPath`, `NewPath`. Same structure as `DeleteNetAppBatchDto`.
- Validator: reuse the rules in `DeleteNetAppBatchRequestValidator` (max 100 operations, not empty, no duplicates, per-item `..` and leading-slash checks) and apply the path rules to both `CurrentPath` and `NewPath`. Add a per-item `CurrentPath != NewPath`, and dedupe on `NewPath` too, since two items renaming to the same destination is a guaranteed conflict you can reject up front.
- Response: same shape as `DeleteNetAppBatchResponse` (overall `Status`, `TotalRequested`, counts, and a per-item `Results` list), with `Conflict` as a first-class per-item outcome next to `NotFound`.
- Function: copy the `DeleteNetAppBatch.Run` structure. Validate, load `caseMetadata`, build `casePrefix`, resolve the security group, then loop with a per-item try/catch, scope both paths to the case folder the way delete does, collect results, write one batch activity log, and always return 200 with per-item statuses.

One dependency to flag: the batch loop can't work until [this is fixed](https://github.com/CPS-Innovation/Large-and-Complex-Cases/pull/469/changes#r3435618550). `DeleteNetAppBatch` only survives a bad item because `DeleteFileOrFolderAsync` returns a result instead of throwing on expected outcomes. As `RenameMaterialAsync` stands, the real `HttpResponseHandler` throws on a 404 or 409, so the first not-found or conflict item would throw straight out of the loop and abort the rest of the batch. Either give the rename client a result-returning shape like the delete client (preferred, matches the pattern most closely), or wrap each item's call in a `try/catch` for `OntapNotFoundException` and `OntapConflictException` inside the loop.

NeilFoubister: @HasanCPS are we including `Type` in the `RenameNetAppBatchOperationDto` for the same purpose as `DeleteNetAppBatchDto` so that we operate on files and/or folders. Therefore, include operations on folders in this ticket?

HasanCPS: `Type` came from copying the delete DTO, not because rename needs it. In delete, `Type` does real work: folders delete recursively and need a trailing slash, files don't. Rename has no such split. From the code it's a single `PATCH .../files/{path}` whether the target is a file or a folder.

The only places `Type` would still help on rename are validation (folders end in `/`) and the activity log (so we can label it folder vs material the way delete does).

So `Type` doesn't decide whether folder rename is in scope. That's a ticket question. Two options:

- Folders in scope: keep `Type`, mirror delete's validation and audit split. First confirm ONTAP renames a directory through the same files endpoint, worth a quick test on the cluster.
- Not in scope: drop `Type`, validate file-only rename, add it later.

backend/CPS.ComplexCases.API/Functions/MaterialBatchRename.cs
```diff
@@ -0,0 +1,87 @@
...
+        _initializationHandler.Initialize(context.Username, context.CorrelationId, renameRequest.Value.CaseId);
+
+        var securityGroups = await _securityGroupMetadataService.GetUserSecurityGroupsAsync(context.BearerToken);
+
+        var result = await _ontapHttpClient.RenameMaterialAsync(
+            context.BearerToken,
+            securityGroups.First().VolumeUuid,
+            renameRequest.Value.CurrentPath,
+            renameRequest.Value.NewPath);
+
```
HasanCPS, line 91: ### Rename isn't confined to the connected case

Nothing ties the rename to the case it claims to belong to. The only thing scoping the operation is the security group's `VolumeUuid` (line 68, `securityGroups.First().VolumeUuid`), and that volume is per security group, not per case. A single volume holds many cases' folders.

`CaseId` does nothing protective. It's used in exactly two places, neither of which constrains the path: line 62 passes it to `_initializationHandler.Initialize` for logging context, and line 80 writes it into the activity log. The function never loads `caseMetadata` or `NetappFolderPath`, so there's no `casePrefix` check, and `CurrentPath` / `NewPath` go straight from the request body to ONTAP. Compare `DeleteNetAppBatch.cs`, which pulls `caseMetadata.NetappFolderPath`, builds a `casePrefix`, and rejects any op whose path isn't under it.

Two consequences:

- A caller can rename material under a different case inside the same volume.
- They can send any `CaseId` they like, so the activity log can record a `CaseId` that doesn't match the path that was actually renamed, which corrupts the audit trail.

The missing `..` and leading-slash guards in the validator (see the validator comment below) make this worse, since a path isn't even confined to its intended subtree.

Fix mirrors delete: load the case's `NetappFolderPath`, derive a `casePrefix`, and reject the rename unless both `CurrentPath` and `NewPath` sit under it. That also makes `CaseId` meaningful instead of decorative.

NeilFoubister: CaseId was included for the logging, so not purely decorative. Point taken though.

backend/CPS.ComplexCases.API/Functions/MaterialRename.cs
```diff
@@ -0,0 +1,231 @@
...
+                    _logger.LogInformation("Failed to rename material from {CurrentPath} to {NewPath}. Error: {ErrorMessage}", operation.CurrentPath, operation.NewPath, result.ErrorMessage);
+                    results.Add(new MaterialRenameBatchItemResult
+                    {
+                        PreviousPath = operation.CurrentPath,
+                        NewPath = operation.NewPath,
+                        Status = OperationResultStatus.Failed,
+                        Error = result.ErrorMessage
+                    });
+                }
+                else if (!result.WasFound)
```
HasanCPS: ### The `NotFound` branch never runs, so a 404 is reported as `Failed`

`RenameMaterialAsync` returns these shapes:

- 200 → `(Success: true,  WasFound: true)`
- 404 → `(Success: false, WasFound: false)`
- 409 → `(Success: false, WasFound: true)`

The function checks `if (!result.Success)` first, which is true for both 404 and 409, so a not-found item lands in the `Failed` branch with error `"Material not found at path: ..."`. The `else if (!result.WasFound)` on line 126 is only reached when `Success` is true, and in that case `WasFound` is always true, so it never fires.

Consequences:

- `NotFound` count is always 0.
- The `NoOp` batch status (documented as "all items NotFound") can never occur; an all-not-found batch reports `Failed`.
- The audit gate `succeeded > 0 || notFound > 0` skips the activity log for an all-not-found batch, even though not-found was meant to be an audited outcome.

Reorder so the happy path is checked first, then `WasFound` separates not-found from real failures:

```csharp
if (result.Success)        { /* Renamed */ }
else if (!result.WasFound) { /* NotFound (404) */ }
else                       { /* Failed (conflict/other) */ }
```

No test caught this: `MaterialRenameTests` only mocks the success result `(true, true)`. Add a case where the client returns `(false, false)` and assert `response.NotFound == 1`.

backend/CPS.ComplexCases.API/Validators/Requests/MaterialRenameRequestValidator.cs
```diff
@@ -0,0 +1,72 @@
...
+                .Must(path => !path.Contains(".."))
+                .WithMessage("NewPath cannot contain '..' to navigate up directories.");
+
+            op.RuleFor(x => x.CurrentPath)
+                .Must(path => !path.StartsWith('/'))
+                .WithMessage("CurrentPath cannot start with a '/'.");
+
+            op.RuleFor(x => x.CurrentPath)
+                .Must(path => !path.StartsWith('/'))
+                .WithMessage("CurrentPath cannot start with a '/'.");
```
HasanCPS: ### Duplicated rule

The `CurrentPath cannot start with a '/'` rule is declared twice; the second block is a copy-paste. Delete one.

backend/CPS.ComplexCases.API/Validators/Requests/MaterialRenameRequestValidator.cs
```diff
@@ -0,0 +1,72 @@
...
+
```
HasanCPS: ### Still no `CurrentPath != NewPath` rule

A no-op rename still round-trips to ONTAP. Add a per-item rule.

backend/CPS.ComplexCases.API/Functions/MaterialRename.cs
```diff
@@ -0,0 +1,231 @@
...
+                !operation.NewPath.StartsWith(casePrefix, StringComparison.OrdinalIgnoreCase))
+            {
+                _logger.LogWarning(
+                    "Either the current path {CurrentPath} or the new path {NewPath} is not within case folder {CasePrefix}. Skipping.",
+                    operation.CurrentPath, operation.NewPath, casePrefix);
+                results.Add(new MaterialRenameBatchItemResult
+                {
+                    PreviousPath = operation.CurrentPath,
+                    NewPath = operation.NewPath,
+                    Status = "Failed",
```
HasanCPS: ### Use the constant, not a literal

The path-scope-skip branch sets `Status = "Failed"` as a string literal while every other branch uses `OperationResultStatus.Failed`. The value matches so counting works, but use the constant.

backend/CPS.ComplexCases.API/Domain/Response/MaterialRenameBatchResponse.cs
```diff
@@ -0,0 +1,27 @@
+namespace CPS.ComplexCases.API.Domain.Response;
+
+public class MaterialRenameBatchResponse
+{
+    /// <summary>
+    /// Overall outcome of the batch.
+    /// Completed: all items processed with no failures (some may be NotFound).
+    /// PartiallyCompleted: at least one item succeeded and at least one failed.
+    /// Failed: no items succeeded; all that were attempted resulted in failure.
+    /// NoOp: no items needed deleting (all were NotFound).
```
HasanCPS: ### Doc comment copied from delete

The `<summary>` says "NoOp: no items needed **deleting**". Should say renaming.

backend/CPS.ComplexCases.API/Functions/MaterialRename.cs
```diff
@@ -0,0 +1,231 @@
...
+    [OpenApiResponseWithBody(statusCode: HttpStatusCode.OK, contentType: ContentType.ApplicationJson, bodyType: typeof(MaterialRenameBatchResponse), Description = ApiResponseDescriptions.Success)]
+    [OpenApiResponseWithBody(statusCode: HttpStatusCode.BadRequest, contentType: ContentType.TextPlain, typeof(string), Description = ApiResponseDescriptions.BadRequest)]
+    [OpenApiResponseWithBody(statusCode: HttpStatusCode.Unauthorized, contentType: ContentType.TextPlain, typeof(string), Description = ApiResponseDescriptions.Unauthorized)]
+    [OpenApiResponseWithBody(statusCode: HttpStatusCode.Forbidden, contentType: ContentType.TextPlain, typeof(string), Description = ApiResponseDescriptions.Forbidden)]
+    [OpenApiResponseWithBody(statusCode: HttpStatusCode.InternalServerError, contentType: ContentType.TextPlain, typeof(string), Description = ApiResponseDescriptions.InternalServerError)]
+    public async Task<IActionResult> Run([HttpTrigger(AuthorizationLevel.Anonymous, "patch", Route = "v1/material/rename")] HttpRequest req, FunctionContext functionContext)
```
HasanCPS: ## Batch endpoint doesn't follow the /batch convention

The function takes a list of operations and returns a batch response, but the route and verb break from its siblings:

| Function | Route | Verb |
|---|---|---|
| DeleteNetAppBatch | `v1/netapp/delete/batch` | POST |
| InitiateBatchMove | `v1/netapp/move/batch` | POST |
| InitiateBatchCopy | `v1/netapp/copy/batch` | POST |
| **MaterialRename** | `v1/material/rename` | PATCH |

Different prefix (`v1/material` vs `v1/netapp`), different verb (PATCH vs POST), and no `/batch` suffix despite being a batch. To match the rest this wants to be `POST v1/netapp/rename/batch`. The class name `MaterialRename` (singular) also reads as a single-item op when it isn't.

backend/CPS.ComplexCases.API/Functions/MaterialRename.cs
```diff
@@ -0,0 +1,231 @@
...
+                var result = await _ontapHttpClient.RenameMaterialAsync(arg);
+
+                if (result.Success)
+                {
+                    results.Add(new MaterialRenameBatchItemResult
+                    {
+                        PreviousPath = operation.CurrentPath,
+                        NewPath = operation.NewPath,
+                        Status = OperationResultStatus.Renamed,
+                        KeysRenamed = result.KeysRenamed > 1 ? result.KeysRenamed : null
```
HasanCPS: ##  KeysRenamed is dead in the response

Surfaces `KeysRenamed` only when `result.KeysRenamed > 1`, but `OntapHttpClient.RenameMaterialAsync` hardcodes `1` on the OK path (`OntapHttpClient.cs:36`). The value can never exceed 1, so `KeysRenamed` is always `null` in every response and the field on `MaterialRenameBatchItemResult` never carries data. Either populate it from the ONTAP response properly, or drop the field and the `> 1` branch.

backend/CPS.ComplexCases.API/Functions/MaterialBatchRename.cs
```diff
@@ -0,0 +1,231 @@
...
+            {
+                _logger.LogError(ex, "Unexpected error renaming material from {CurrentPath} to {NewPath}.", operation.CurrentPath, operation.NewPath);
+                results.Add(new MaterialRenameBatchItemResult
+                {
+                    PreviousPath = operation.CurrentPath,
+                    NewPath = operation.NewPath,
+                    Status = OperationResultStatus.Failed,
+                    Error = ex.Message
+                });
+            }
```
HasanCPS, line 161: ##  Auth failures return 200 OK with everything marked Failed

The per-item `catch (Exception ex)` swallows everything, including `OntapUnauthorizedException`. With a bad/expired bearer token every item throws 401, gets caught, and the batch returns `200 OK` with all items `Failed`. The middleware mapping for `OntapUnauthorizedException -> 401` (`ExceptionHandlingMiddleware.cs:85`) never fires because the exception never leaves the loop. Let whole-batch auth/forbidden failures propagate so the caller gets a real 401/403 instead of a 200 that looks like a per-item data problem.

backend/CPS.ComplexCases.API/Functions/MaterialRename.cs
```diff
@@ -0,0 +1,231 @@
...
+    private readonly ICaseMetadataService _caseMetadataService = caseMetadataService;
+    private readonly IOntapArgFactory _ontapArgFactory = ontapArgFactory;
+    private readonly IOntapHttpClient _ontapHttpClient = ontapHttpClient;
+    private readonly FeatureFlagConfig _featureFlags = featureFlags.Value;
+
+    [Function(nameof(MaterialRename))]
+    [OpenApiOperation(operationId: nameof(MaterialRename), tags: ["Material"], Description = "Rename a material folder for a case.")]
+    [CmsAuthValuesAuth]
+    [BearerTokenAuth]
+    [OpenApiRequestBody(ContentType.ApplicationJson, typeof(MaterialRenameRequestDto), Description = "Body containing the material rename information")]
```
HasanCPS: ## Two rename endpoints are live at once — this is the swagger sourcePath/destinationPath

**`CPS.ComplexCases.API/Functions/Transfer/RenameNetAppMaterial.cs:31-44`**

The old single-item endpoint was never removed. It's still registered:

- Route `POST v1/netapp/materials/rename`, tag `NetApp`
- Body `RenameNetAppMaterialRequest` with `sourcePath` / `destinationPath` (`CPS.ComplexCases.API/Domain/Request/RenameNetAppMaterialRequest.cs:10-14`)
- Still backed by the copy-then-delete `FileTransferClient` path, plus its validator `RenameNetAppMaterialRequestValidator.cs`

Swagger now shows two ways to rename. The one displaying `sourcePath`/`destinationPath` is this old function, not the new `MaterialRename` (which correctly uses `caseId` + `operations[].currentPath/newPath`).

**Action: remove the old endpoint, keep the new one.** The new ONTAP batch rename replaces the old copy-then-delete approach, so the old function is now redundant. Leaving both live means two divergent rename APIs with different request shapes in swagger, which will confuse anyone integrating against it. Delete:

- `RenameNetAppMaterial.cs` (the function)
- `RenameNetAppMaterialRequest.cs` (the request DTO with sourcePath/destinationPath)
- `RenameNetAppMaterialRequestValidator.cs` (its validator)
- the `RenameNetAppMaterialAsync` call path on `FileTransferClient` if nothing else uses it

Keep `MaterialRename` as the single source of truth.

backend/CPS.ComplexCases.Common/Models/Configuration/FeatureFlagConfig.cs
```diff
@@ -8,4 +8,9 @@ public class FeatureFlagConfig
     /// Enables case_active_manage_materials table access on GetCase API.
     /// </summary>
     public bool ManageMaterials { get; set; } = false;
+
+    /// <summary>
+    /// Enables the material rename functionality.
+    /// </summary>
+    public bool MaterialRename { get; set; } = false;
```
HasanCPS, line 15: ## The `MaterialRename` feature flag is never set in the DevOps pipeline

The new `MaterialRename` flag defaults to `false`, and `MaterialBatchRename.cs:62` returns a 404 whenever it's off. The problem is that nothing in `devops-pipelines/` ever turns it on, so it stays `false` in every deployed environment and the endpoint is effectively dead on arrival.

Compare with the existing `ManageMaterials` flag, which is wired in two places:

- `devops-pipelines/templates/fa-config-steps.yml:223` maps the app setting `FeatureFlags__ManageMaterials` to `$(FeatureFlagsManageMaterials)`. There is no matching block for `FeatureFlags__MaterialRename`.
- `fa-config-dev.yml:40`, `fa-config-staging.yml:40` and `fa-config-prod.yml:40` each declare a `FeatureFlagsManageMaterials` variable. None of them declare a `MaterialRename` equivalent.

To ship this we need both: an app-setting block in `fa-config-steps.yml` and a `FeatureFlagsMaterialRename` variable in each of the three `fa-config-<env>.yml` files. Otherwise the feature can't be switched on without a code change.

devops-pipelines/templates/variables/fa-config-dev.yml
```diff
@@ -39,3 +39,5 @@ variables:
     value: "10000"
   - name: FeatureFlagsManageMaterials
     value: "true"
+  - name: FeatureFlagsMaterialRename
+    value: "false"
```
HasanCPS: Enable this setting for development and staging environments (set to **true**), and disable it in production (set to **false**).

backend/CPS.ComplexCases.NetApp/Extensions/IServiceCollectionExtension.cs
```diff
@@ -105,6 +112,25 @@ public static void AddNetAppClient(this IServiceCollection services, IConfigurat
 			return _netAppS3HttpClientPolicy.Value;
 		});
 
+		services.AddHttpClient<IOntapHttpClient, OntapHttpClient>(client =>
```
HasanCPS, line 115: ## ONTAP client timeout and resilience wiring belong in FCT2-19919

The new ONTAP HTTP client registration brings in two things that another branch is actively reworking, so I'd rather not settle them here.

First, the timeout. The ONTAP client sets `client.Timeout = TimeSpan.FromMinutes(10)` directly, while the two existing NetApp clients in the same method read it from config: `configuration.GetValue("NetAppOptions:RequestTimeoutSeconds", 100)`. One client out of step with the other two.

Second, the resilience handler. The ONTAP client adds a new static `_ontapHttpClientPolicy` `Lazy` field and the `.AddPolicyHandler((sp, _) => { _serviceProvider ??= sp; return _ontapHttpClientPolicy.Value; })` block, copying the pattern the existing clients use.

Here's why both should move to FCT2-19919 (`feature/fct2_19919_object_disposed_exception`) rather than land here:

- That branch already carries the FCT2-19532 timeout work (it's based on top of PR #479), so the config-driven `RequestTimeoutSeconds` timeout is already in it.
- More importantly, 19919 deletes the static `_serviceProvider` and `Lazy<...>` policy fields and swaps every `.AddPolicyHandler((sp, _) => { _serviceProvider ??= sp; ... })` for a clean `.AddResiliencePolicyHandler(GetResiliencePolicy)`. That static captured-provider pattern is the ObjectDisposedException it's fixing. The ONTAP client here re-introduces the exact field and block that branch is removing, so it'll both collide on this file and put the disposed-provider risk back for the ONTAP client.

The resiliency policy itself is fine, to be clear. `GetResiliencePolicy` reuses `HttpResiliencePolicyFactory` from `Common/Resilience/`, which came in with FCT2-19531 (circuit breakers, PR #473, already on main). No duplication.

Since this PR is already growing and 19919 (PR #481) rewrites this exact block, suggest we fold the ONTAP client into 19919: migrate it to `.AddResiliencePolicyHandler` and the config-driven timeout there, in one place, instead of adding a copy here that 19919 then has to unpick. I suggest @bridgesr and @NeilFoubister work together on this.

HasanCPS (approved):
LGTM!

PR #470, opened 2026-06-15
Story/FCT2 - 17260 Transfer Material New Design

renjithabby:
https://cpsgovuk.atlassian.net/browse/FCT2-17260

to locally test please add VITE_FEATURE_FLAG_TRANSFER_MATERIALS_V1=true to .env.local


ui-spa/src/apis/gateway-api.ts
```diff
@@ -325,13 +325,20 @@ export const getCaseMetaData = async (caseId: string) => {
...
+  folderParamKey: "folder-id" | "path" = "folder-id",
   skip: number = 0,
   take: number = 50,
   collected: EgressFolderData = [],
 ): Promise<EgressFolderData> => {
+  let folderParam: { "folder-id"?: string } | { path?: string } = {
+    "folder-id": folderIdOrPath,
+  };
+
+  if (folderParamKey === "path")
```
HasanCPS, line 338: ## Merge latest main: this branch predates the egress `path` support it depends on

**File:** `ui-spa/src/apis/gateway-api.ts` (the new `folderParamKey: "path"` branch), called from `ui-spa/src/components/case-management/transfer-materials-v1/index.tsx:149`

The v1 source table now calls `getEgressFolders(egressWorkspaceId, egressFolderPath, "path")`, which sends `?path=<folderPath>` and drops `folder-id`. That relies on the egress list endpoint understanding `path`.

The backend support exists on `main` already, added by PR #471 ("enhance file transfer response and add path to egress listing"). On main, `ListEgressMaterials.cs` reads `path` (line 48), derives `pathArg = string.IsNullOrEmpty(folderId) ? path : null` (line 53), and passes it to `CreateListWorkspaceMaterialArg(...)` (line 55). The frontend sends `path` only with no `folder-id`, so `pathArg` resolves to the path, which is what we want.

This branch's merge-base is behind #471, so the branch is carrying the old `folder-id`-only version of `ListEgressMaterials.cs`. Please merge latest `main` so the egress endpoint actually consumes `path`. Without it, egress navigation passes CI (mocks honour `path`) but returns the root listing against the real backend.

After merging, double-check two contract details line up:
- the frontend strips the trailing slash (`folderIdOrPath.replace(/\/$/, "")`) before sending `path`, so confirm the endpoint expects the path without a trailing slash.
- main's endpoint gives `folder-id` precedence over `path`. The frontend omits `folder-id` for the path call, so this is fine, just worth a glance after the merge.

ui-spa/src/components/case-management/transfer-materials-v1/index.tsx
```diff
@@ -0,0 +1,793 @@
...
+        .map((data) => ({
+          fileId: data.id,
+          path: data.path,
+          isFolder: data.isFolder,
+        }));
+    } else {
+      return selectedSourceFoldersOrFiles.map((path) => ({
+        path: path,
+      }));
+    }
```
HasanCPS, line 490: ## Select-all in NetApp mode ships the `"all-folders"` sentinel as a source path

`handleCheckboxChange` stores the literal `"all-folders"` string as the first entry when select-all is ticked (`index.tsx:420-431`). The egress branch of `getTransferSourcePath` is safe because it filters `egressFolderData` by `includes(...)`, which drops the sentinel. The NetApp branch maps `selectedSourceFoldersOrFiles` straight through:

```ts
return selectedSourceFoldersOrFiles.map((path) => ({ path }));
```

So a select-all copy or move from the Shared Drive sends `{ path: "all-folders" }` to the backend, and it also leaks into `getCommonPath` / `sourceRootFolderPath` on the destination page. Filter out `"all-folders"`, or build the paths from `netAppFolderData` the way the egress branch does.

ui-spa/src/components/case-management/transfer-materials-v1/index.tsx
```diff
@@ -0,0 +1,793 @@
...
+        setActiveTransferData(null);
+      }, 0);
+    } else if (
+      transferSource === "egress" &&
+      activeTransferData?.direction === "NetAppToEgress"
+    ) {
+      setEgressFolderPath(activeTransferData.destinationPath);
+      setActiveTransferData(null);
+    }
+  };
```
HasanCPS, line 555: ## The success banner "go to folder" link does nothing for NetApp to Egress transfers

```ts
if (transferSource === "egress" && direction === "EgressToNetApp") { ... }
else if (transferSource === "egress" && direction === "NetAppToEgress") { ... }
```

During a NetApp to Egress transfer the user's `transferSource` is `"netapp"`, so the second branch (which requires `transferSource === "egress"`) can never run. The destination link in the success banner is dead for that direction. The branch most likely needs to test `transferSource === "netapp"`, then switch the source to egress and set `egressFolderPath`, the same way the first branch switches to netapp.

ui-spa/src/components/case-management/transfer-materials-v1/index.tsx
```diff
@@ -0,0 +1,793 @@
...
+  const [sortValues, setSortValues] = useState<{
+    name: string;
+    type: "ascending" | "descending";
+  }>();
+
+  const handleTableSort = (
+    sortName: string,
+    sortType: "ascending" | "descending",
+  ) => {
+    console.log("Table sorted:", { sortName, sortType });
```
HasanCPS: ## Remove leftover debug logging

**Files:**
- `ui-spa/src/components/case-management/transfer-materials-v1/index.tsx:136`
- `ui-spa/src/components/case-management/transfer-materials-v1/index.tsx:515`
- `ui-spa/src/components/case-management/transfer-materials-v1/TransferDestinationPage.tsx:326`

Three `console.log` calls left in, including one that logs the folder object on every NetApp folder click. Drop them.

ui-spa/src/components/case-management/transfer-materials-v1/utils/getSortedTableData.tsx
HasanCPS, line 1: ## Delete the two unused sort utilities

**Files:**
- `ui-spa/src/components/case-management/transfer-materials-v1/utils/getSortedTableData.tsx`
- `ui-spa/src/components/case-management/transfer-materials-v1/utils/isEgressFolderData.tsx`

Neither is imported anywhere. `index.tsx` does its sorting inline with `sortUtils` instead. `getSortedTableData` also has an `if/else` in its `file-size` branch where both arms return the same call. Either wire these in or remove them so they do not rot.

ui-spa/src/components/case-management/transfer-materials-v1/TransferDestinationPage.tsx
```diff
@@ -0,0 +1,417 @@
...
+          </h1>
+          <p>
+            {sourcePaths.length === 1
+              ? "You are copying 1 item."
+              : `You are copying ${sourcePaths.length} items.`}
+          </p>
+          <p>
+            {transferSource === "egress"
+              ? "Select the Shared Drive folder you want to copy them into."
+              : "Select the Egress folder you want to copy them into."}
```
HasanCPS: ## Destination page says "copying" even for a Move

The heading copy reads "You are copying N items" and "the folder you want to copy them into" regardless of `selectedTransferAction`. When the action is Move, the user still sees "copying". Switch the wording on `selectedTransferAction`.

ui-spa/src/components/case-management/transfer-materials-v1/index.tsx
renjithabby: fixed

ui-spa/src/components/case-management/transfer-materials-v1/index.tsx
renjithabby: fixed

ui-spa/src/apis/gateway-api.ts
renjithabby: i have checked  it , with the latest changes and against deployed dev , looks good to me

ui-spa/src/components/case-management/transfer-materials-v1/TransferDestinationPage.tsx
```diff
@@ -0,0 +1,440 @@
...
+    const paths = sourcePaths.map(({ path }) => path);
+    const sourceCurrentFolderPath = getCommonPath(paths);
+    return {
+      transferSource: transferSource,
+      transferEgressFolderPathInitialValue:
+        transferSource === "egress" ? sourceCurrentFolderPath : undefined,
+      transferNetAppFolderPathInitialValue:
+        transferSource === "netapp" ? sourceCurrentFolderPath : undefined,
+    };
+  }, [sourcePaths, transferSource]);
```
HasanCPS, line 220: ### Back-nav restore lands on `"/"` instead of root when one item is selected at the egress root

`caseManagementRouteState` derives the folder to restore with `getCommonPath(sourcePaths)`. For a single selection, `getCommonPath` returns `getRelativePathFromPath(path) + "/"`. Egress paths are relative, so a single top-level item (e.g. folder `"Documents"` or file `"report.pdf"`) has no parent segment: `getRelativePathFromPath` returns `""`, and `getCommonPath` hands back `"/"`.

On return, `egressFolderPath` initialises to `"/"` rather than `""`. The listing query still strips the trailing slash so it fetches the right folder, but `"/"` is truthy, so the root-only guards keyed on `!egressFolderPath` now read as "not root":
- the checkbox column guard at `index.tsx:236` (`!egressFolderData.length || !egressFolderPath`) stops hiding the column, so the user can now tick boxes at what is effectively the egress root — the exact thing the disabled-root work was meant to prevent;
- the breadcrumb (`getPathFolders(egressFolderPath, ...)`, `index.tsx:114`) gets a stray empty segment.

Multi-select at root is fine (`commonPathPrefix` returns `""`), and NetApp is unaffected because its paths are absolute so there is always a parent. Suggest normalising the restore value, e.g. treat `"/"` (or a path that resolves to the root) as `""` before putting it in the route state, so the restored state matches a fresh root visit.

ui-spa/src/components/case-management/transfer-materials-v1/index.tsx
HasanCPS, line 552: ### `handleGotoFolderClick` has a redundant inner direction check

```ts
if (transferSource === "netapp" && activeTransferData) {
  if (activeTransferData?.direction === "NetAppToEgress") {
    ...
  } else {
    setNetAppFolderPath(activeTransferData.destinationPath);
  }
```

After [90316e4](https://github.com/CPS-Innovation/Large-and-Complex-Cases/pull/470/changes/90316e475a0a65b559ffde08f79f729827ca5bec) loosened the outer guard to `activeTransferData` only, the inner `if` re-checks `direction === "NetAppToEgress"`. The `else` arm sets a NetApp path from a destination that, for a NetApp-sourced transfer, is an egress path — so if it ever runs it does the wrong thing. In practice it is unreachable for a normal NetApp→Egress flow. Either drop the inner check (mirror of the egress branch above it, which keeps the `else` for the source-toggled-mid-transfer case) or keep it deliberately and document why. As written it reads like leftover scaffolding.

ui-spa/src/components/case-management/transfer-materials-v1/TransferDestinationPage.tsx
renjithabby: fixed

ui-spa/src/components/case-management/transfer-materials-v1/index.tsx
renjithabby: fixed

ui-spa/src/components/case-management/transfer-materials-v1/TransferDestinationPage.tsx
```diff
@@ -0,0 +1,440 @@
...
+          }[]
+        | { path: string }[];
+      transferSource: "egress" | "netapp";
+      egressWorkspaceId: string;
+      selectedTransferAction: "copy" | "move";
+      caseId: number;
+      netAppPath: string;
+      operationName: string;
+    };
+  } = useLocation();
```
HasanCPS, line 51: ### Crashes on refresh / direct nav

```ts
const {
  state: { transferSource, sourcePaths, caseId, ... },
}: { state: { ... } } = useLocation();
```

`state` is destructured non-optionally. React Router returns `state: null` on a hard refresh
or direct link, so this throws *"Cannot destructure property 'transferSource' of … as it is
null"* and the page white-screens (hits the error boundary). Every sibling page handles this
defensively — e.g. `netapp-disconnect/DisconnectSharedDriveConfirmationPage.tsx:20`:

```ts
const { state } = useLocation();
const { caseId, urn } = state || {};
```

The destination page is a transient step in the flow, so a mid-flow refresh is a realistic
way to hit it. Suggest reading `state` optionally and redirecting back to
`/case/${caseId}/case-management` (or rendering a fallback) when it's missing, consistent
with the `isRouteValid` convention the rest of the flow uses.

ui-spa/src/common/utils/sortUtils.ts
```diff
@@ -36,8 +45,19 @@ export const sortByNumberProperty = <T>(
...
-    const sizeB = typeof b[numberProperty] === "number" ? b[numberProperty] : 0;
+    let sizeA = a[numberProperty] as number;
+    let sizeB = b[numberProperty] as number;
+
+    if (!sizeA)
+      sizeA =
+        order === "ascending"
+          ? Number.POSITIVE_INFINITY
+          : Number.NEGATIVE_INFINITY;
+    if (!sizeB)
```
HasanCPS: ### `!sizeA` treats a real `0` as "missing"

```ts
let sizeA = a[numberProperty] as number;
if (!sizeA) sizeA = order === "ascending" ? +Infinity : -Infinity;
```

`!0` is `true`, so a genuine 0-byte file is bucketed with folders (whose `filesize` is `0`)
and shoved to the end of the list. That's probably the intent — keep "--" rows at the bottom
— but it conflates "zero" with "absent". If 0-byte files should sort as 0, use
`Number.isFinite(sizeA)` / an explicit nullish check instead. The date-sort version above it
is fine because it keys off `Number.isNaN` only.

ui-spa/src/components/case-management/transfer-materials-v1/index.tsx
```diff
@@ -0,0 +1,811 @@
...
+  const handleFolderClick = (data: EgressFolder | NetAppFileFolder) => {
+    if (transferSource === "egress" && "id" in data) {
+      setEgressFolderPath(data.path);
+      if (transferSource === "egress") setSelectedSourceFoldersOrFiles([]);
+    }
+    if (transferSource === "netapp") {
+      setNetAppFolderPath(data.path);
+      if (transferSource === "netapp") setSelectedSourceFoldersOrFiles([]);
+    }
+  };
```
HasanCPS, line 529: ### Redundant inner guard

```ts
if (transferSource === "egress" && "id" in data) {
  setEgressFolderPath(data.path);
  if (transferSource === "egress") setSelectedSourceFoldersOrFiles([]); // already known true
}
```

The inner `if (transferSource === ...)` checks are always true inside their outer branches.
Simplify to just the `setSelectedSourceFoldersOrFiles([])` calls.

ui-spa/src/components/common/tree-view-component/TreeViewComponent.tsx
```diff
@@ -22,15 +25,19 @@ export type TreeViewComponentProps = {
   onSelect?: (node: TreeNode) => void;
   onLoadChildren?: (nodeId: string) => Promise<TreeNode[]>;
   className?: string;
+  isRootNodeOpened?: boolean;
```
HasanCPS, line 28: ### `isRootNodeOpened` implemented + tested but never wired

The only production consumer, `TransferWidget.tsx:34-38`, doesn't pass `isRootNodeOpened`, so
the destination tree root never auto-expands in the real app (default `false`). Either pass
it from `TransferWidget` if root-open is the desired UX, or drop the prop to avoid a tested-
but-dead code path.

Note also that the initializer `new Set([data[0]?.id])` only runs on mount — if `data` is
empty on first render and fills in later, it won't expand. Fine for the current call sites
(root node is always present synchronously) but a trap if reused.

ui-spa/src/common/utils/getCommonPath.test.ts
```diff
@@ -29,4 +29,13 @@ describe("getCommonPath", () => {
...
   });
+  it("Should return empty string as relative path if the input paths array is empty", () => {
+    expect(getCommonPath(["", ""])).toEqual("");
+  });
+  it("Should return empty string as relative path if the input paths array contains only one empty string path", () => {
+    expect(getCommonPath([""])).toEqual("");
+  });
+  it("Should return empty string as relative path if the input paths array contains more than one empty string path", () => {
+    expect(getCommonPath(["", ""])).toEqual("");
+  });
```
HasanCPS, line 40: ### Misleading + duplicate cases (carried over)

First case description says "array is empty" but passes `["", ""]`; cases 1 and 3 are
identical. Swap one for `[""]` (the branch the fix actually added) and, if you want, cover
`getCommonPath([])`.

renjithabby: fixed

ui-spa/src/components/common/tree-view-component/TreeViewComponent.tsx
renjithabby: fixed

ui-spa/src/components/case-management/transfer-materials-v1/index.tsx
renjithabby: fixed

ui-spa/src/common/utils/sortUtils.ts
renjithabby: fixed

ui-spa/src/components/case-management/transfer-materials-v1/TransferDestinationPage.tsx
renjithabby: fixed

HasanCPS (approved):
LGTM!

PR #471, opened 2026-06-16
enhance file transfer response and add path to egress listing

bridgesr:


backend/CPS.ComplexCases.API/Functions/ListEgressMaterials.cs
```diff
@@ -43,10 +44,14 @@ public async Task<IActionResult> Run([HttpTrigger(AuthorizationLevel.Anonymous,
...
         var folderId = req.Query[InputParameters.FolderId];
+        var path = req.Query[InputParameters.Path];
         var skip = int.TryParse(req.Query[InputParameters.Skip], out var skipValue) ? skipValue : 0;
         var take = int.TryParse(req.Query[InputParameters.Take], out var takeValue) ? takeValue : 100;
 
-        var listMaterialsArg = _egressArgFactory.CreateListWorkspaceMaterialArg(workspaceId, skip, take, folderId, null);
+        // folder-id takes precedence over path when both are supplied.
+        var pathArg = string.IsNullOrEmpty(folderId) ? path.FirstOrDefault() : null;
+
+        var listMaterialsArg = _egressArgFactory.CreateListWorkspaceMaterialArg(workspaceId, skip, take, folderId, null, pathArg);
```
HasanCPS:
```csharp
var folderId = req.Query[InputParameters.FolderId];   // StringValues -> string, comma-joins if repeated
var path = req.Query[InputParameters.Path];
...
var pathArg = string.IsNullOrEmpty(folderId) ? path.FirstOrDefault() : null;
```

`folderId` rides the implicit `StringValues → string` conversion (which comma-joins repeated params), while `path` uses `.FirstOrDefault()`. Harmless since neither param is expected to repeat, but the asymmetry reads oddly. Pick one approach.

backend/CPS.ComplexCases.API/Functions/ListEgressMaterials.cs
HasanCPS, line 42: The `{workspaceId}` route segment isn't declared as an OpenAPI parameter, so Swagger UI renders no input box for it and the endpoint can't be exercised from the docs page. The query params (`folder-id`, `path`, `skip`, `take`) each have an `[OpenApiParameter]`, but the route path param doesn't — the OpenAPI generator only documents parameters it's explicitly given. The HTTP route itself still works; this only affects the generated spec / Swagger UI.

This is the only function in the API with a `{workspaceId}` route, which is why it's easy to miss.

Add the path-parameter attribute, matching the `caseId` convention used in `GetCase`, `ProvisionNetAppFolders`, etc.:

```csharp
[BearerTokenAuth]
[OpenApiParameter(name: "workspaceId", In = ParameterLocation.Path, Required = true, Type = typeof(string), Description = "The Egress workspace ID to list files/folders for.")]
[OpenApiParameter(name: InputParameters.FolderId, In = ParameterLocation.Query, ...)]

HasanCPS (approved):
LGTM!

PR #472, opened 2026-06-16
update pipeline templates to set ASPNETCORE_ENVIRONMENT

lilachdavis:
- update fa-config-steps to set a value for `ASPNETCORE_ENVIRONMENT`
- update per-env variables to include the safe values (PreProd / Production)

Not in scope but I thought it made sense to add this too:
- update same as above with the env var `AZURE_FUNCTIONS_ENVIRONMENT`


HasanCPS (approved):
LGTM!

PR #473, opened 2026-06-17
Feature/fct2 19531 circuit breakers

bridgesr:


backend/CPS.ComplexCases.DDEI/Extensions/IServiceCollectionExtension.cs
```diff
@@ -19,13 +19,23 @@ public static class IServiceCollectionExtension
...
+  private static IAsyncPolicy<HttpResponseMessage>? _resiliencePolicy;
+
   public static void AddDdeiClient(this IServiceCollection services, IConfiguration configuration)
   {
     services.Configure<DDEIOptions>(configuration.GetSection(nameof(DDEIOptions)));
     services.AddTransient<IDdeiArgFactory, DdeiArgFactory>();
     services.AddHttpClient<IDdeiClient, DdeiClient>(AddDdeiClient)
       .SetHandlerLifetime(TimeSpan.FromMinutes(5))
-      .AddPolicyHandler(GetRetryPolicy());
+      .AddPolicyHandler((sp, _) => _resiliencePolicy ??= GetResiliencePolicy(sp.GetRequiredService<ILoggerFactory>()));
```
HasanCPS: ###  `??=` lazy init is not thread-safe

`backend/CPS.ComplexCases.DDEI/Extensions/IServiceCollectionExtension.cs:38`
(same pattern at `CPS.ComplexCases.Egress/.../IServiceCollectionExtension.cs:44` and `:57`,
and `CPS.ComplexCases.NetApp/.../IServiceCollectionExtension.cs:79` and `:95`)

```csharp
.AddPolicyHandler((sp, _) => _resiliencePolicy ??= GetResiliencePolicy(sp.GetRequiredService<ILoggerFactory>()));
```

The policy selector runs per request. Under a burst of concurrent first requests, two threads can
both find the field null and both build a policy. The reference assignment will not tear, so this
settles to one instance straight away, but for a moment you can have two breaker instances with
split state. Low impact. If you want it airtight, a `Lazy<IAsyncPolicy<HttpResponseMessage>>` per
client says "one shared instance" without the race.

backend/CPS.ComplexCases.Common/Resilience/HttpResiliencePolicyFactory.cs
```diff
@@ -0,0 +1,99 @@
...
+      ILogger logger,
+      string serviceName,
+      double failureThreshold,
+      TimeSpan samplingDuration,
+      int minimumThroughput,
+      TimeSpan durationOfBreak)
+  {
+    return Policy<HttpResponseMessage>
+        .Handle<HttpRequestException>()
+        .OrResult(r => r.StatusCode >= HttpStatusCode.InternalServerError)
```
HasanCPS, line 41: ### Breaker trips on POST/PUT 5xx while retry does not (confirm this is intended)

`backend/CPS.ComplexCases.Common/Resilience/HttpResiliencePolicyFactory.cs:41` (breaker handles any
5xx, no method filter), versus the retry's `ExcludesPostAndPut` use at `:74`

The retry path excludes POST and PUT (`ExcludesPostAndPut`) because they are not idempotent. The
breaker does not apply that filter, so a 5xx on a POST or PUT still counts toward opening the
circuit. I think that is correct: the breaker should reflect service health regardless of method,
and once it opens it protects every call. Flagging it only so the asymmetry is a decision rather
than an accident.

HasanCPS (approved):
LGTM!

PR #474, opened 2026-06-17
Defect/fct2 19491 investigate grpc issue with move functionality in staging

lilachdavis:
Move transfers were failing intermittently in staging when activities
signalled or read the transfer entity. The Durable Task management calls
threw gRPC StatusCode.Unavailable and underlying WSAEACCES socket errors
under local port pressure on the Windows hosts.

Resolve the DurableTaskClient per activity via the [DurableClient] binding
instead of the singleton registered in Program.cs, and route the entity
calls through a shared DurableEntityRetry helper that retries transient
Unavailable RpcExceptions and SocketExceptions (3 attempts, 1s apart).

- Add DurableEntityRetry helper for transient gRPC/socket failures
- Pass DurableTaskClient into TransferEntityHelper.GetTransferEntityAsync
  and DeleteMovedItemsCompleted
- Bind [DurableClient] in DeleteFiles, FinalizeTransfer and
  UpdateTransferStatus activities
- Remove the AddDurableTaskClient(UseGrpc) registration from Program.cs
- Drop the ad-hoc retry helper that lived in DeleteFiles
- Update unit tests for the new signatures


backend/CPS.ComplexCases.FileTransfer.API/Durable/Helpers/TransferEntityHelper.cs
```diff
@@ -3,28 +3,66 @@
...
+                "Got Durable entity {EntityName}/{EntityKey} for TransferId={TransferId}. EntityFound={EntityFound}, HasState={HasState}",
+                entityId.Name,
+                entityId.Key,
+                transferId,
+                entity is not null,
+                entity?.State is not null);
+
+            return entity;
+        }
+        catch (Exception ex)
```
HasanCPS: `GetTransferEntityAsync` wraps every failure, transient or not, in `InvalidOperationException` (line 56-60), but `DeleteMovedItemsCompleted` (line 23-27) lets the original exception propagate. If anything upstream branches on exception type, that split is going to surprise someone. Wrap both or neither.

backend/CPS.ComplexCases.FileTransfer.API/Durable/Activity/FinalizeTransfer.cs
```diff
@@ -13,6 +15,10 @@ public async Task Run([ActivityTrigger] FinalizeTransferPayload finalizeTransfer
     {
         var entityId = new EntityInstanceId(nameof(TransferEntityState), finalizeTransferPayload.TransferId.ToString());
 
-        await client.Entities.SignalEntityAsync(entityId, nameof(TransferEntityState.FinalizeTransfer), cancellationToken);
+        await DurableEntityRetry.ExecuteAsync(
+            nameof(FinalizeTransfer),
+            () => client.Entities.SignalEntityAsync(entityId, nameof(TransferEntityState.FinalizeTransfer), cancellationToken),
+            NullLogger.Instance,
```
HasanCPS: ### `Durable/Activity/FinalizeTransfer.cs:21` and `Durable/Activity/UpdateTransferStatus.cs:21`
Both pass `NullLogger.Instance` into `DurableEntityRetry`, so the "transient gRPC failure, retrying" warning is dropped for the two operations you just added retries to. These are activities, so inject `ILogger<T>` and pass it through. Otherwise the retries are invisible in staging, which is exactly where we're chasing this.

backend/CPS.ComplexCases.FileTransfer.API/host.json
```diff
@@ -18,7 +18,9 @@
       "Azure.Core": "Warning",
       "Azure.Identity": "Warning",
       "Worker": "Warning",
-      "Grpc.Net.Client": "Warning",
+      "Grpc.Net.Client": "Information", 
+      "Microsoft.DurableTask": "Information",
+      "Microsoft.Azure.Functions.Worker.Extensions.DurableTask": "Information",
```
HasanCPS: `Grpc.Net.Client`, `Microsoft.DurableTask`, and the `Microsoft.Azure.Functions.Worker.Extensions.DurableTask` extension are all raised to `Information`. That helps while you're diagnosing staging, but it's noisy and costly to leave on in prod, so track it to dial back once the fix is confirmed.

HasanCPS (approved):
LGTM!

PR #475, opened 2026-06-18
polling on 404s issue

bridgesr:


ui-spa/src/common/utils/pollTransferStatus.ts
```diff
@@ -5,6 +5,17 @@ import { ApiError } from "../../common/errors/ApiError";
...
+/**
+ * Upper bound on consecutive 404s tolerated before giving up. A 404 means the
+ * Durable entity TransferEntityState/{transferId} does not exist yet. This is
+ * expected for a short window between InitiateTransfer returning Accepted and
+ * the orchestrator creating the entity, so we allow a grace window. Past this
+ * cap we treat the transfer as missing and surface an error instead of polling
+ * forever (e.g. an orchestration that failed before the entity was created, or
+ * a stale/incorrect transfer ID).
+ */
+const MAX_CONSECUTIVE_NOT_FOUND = 12;
```
HasanCPS: **Worth firming up the 404 grace window before we ship**

The cap of 12 consecutive 404s works out to ~11s of wall-clock in production: during a
run of 404s `adaptiveBase` never moves off the 1000ms default (it's only reassigned in the
`data !== null` branch) and `noChangeCount` stays 0, so each retry is a flat 1s and the 12th
poll trips the cap with no trailing delay. So the real budget is 11 polls × 1s plus round-trips,
roughly 11–14s.

That grace window only has to cover the gap between `InitiateTransfer` returning `Accepted`
and the orchestrator creating `TransferEntityState/{transferId}`. We're on an Azure Functions
Premium plan with always-ready instances, so the warm baseline gap is well under 11s and this
is fine for the normal case. Two cases can still run long though, and Premium doesn't remove
either:

- scale-out cold starts — a new instance brought up under load still cold-starts the isolated worker
- Durable control-queue dispatch latency under load

When that happens, a perfectly healthy transfer 404s for the whole window and the user gets a
false "Transfer not found". That false negative is worse than a bit of extra polling, so I'd
bias the budget longer.

Two suggestions, either is fine:

1. (preferred) Make the window time-based rather than a count. Right now "12" only means ~11s
   because the interval happens to be 1000ms — change the default `pollingInterval` and the
   grace window silently changes with it (in the tests at 100ms, "12" is 1.2s). Tracking a
   first-404 timestamp says what we mean and is robust to interval changes:

   ```ts
   let firstNotFoundAt: number | null = null;
   // in the 404 branch:
   firstNotFoundAt ??= Date.now();
   if (Date.now() - firstNotFoundAt >= NOT_FOUND_GRACE_MS) { handleError(...); break; }
   // reset on any 200/304: firstNotFoundAt = null;
   ```

   I'd set `NOT_FOUND_GRACE_MS` to ~30s to absorb scale-out and dispatch latency.

2. If we'd rather keep the count, bump it to ~30 so the budget clears a scale-out cold start,
   and add a comment that the count is interval-coupled.

HasanCPS (approved):
LGTM!

PR #476, opened 2026-06-18
use new KV name for FA config

lilachdavis:
Due to a requirement to increase soft-delete retention on the Prod key vault to 90 days, new KVs had to be created (retention policy cannot be changed after creation).
This is updating any references to the KV to use its new name.


HasanCPS (approved):
LGTM!

PR #479, opened 2026-06-19
Feature/fct2 19532 http timeouts

bridgesr:


backend/CPS.ComplexCases.Egress/Extensions/IServiceCollectionExtension.cs
```diff
@@ -58,7 +58,11 @@ public static void AddEgressClient(this IServiceCollection services, IConfigurat
         throw new ArgumentNullException(nameof(egressServiceUrl), "EgressOptions:Url configuration is missing or empty.");
       }
       client.BaseAddress = new Uri(egressServiceUrl);
-      client.Timeout = TimeSpan.FromMinutes(10);
+      // EgressStorageClient mixes short management calls with file streamed downloads and
+      // chunk uploads. The body read of a streamed download is bound by HttpClient.Timeout, so a
+      // single short value would break large downloads. Per-operation timeouts are enforced inside
+      // BaseEgressClient via CancellationToken instead, leaving the client itself uncapped.
+      client.Timeout = Timeout.InfiniteTimeSpan;
```
HasanCPS, line 65: **Egress download body read has no real timeout now**

Tracing the streamed download path, I think this PR drops the ceiling that used to guard a stalled download, and nothing transfer-sized replaces it.

A streamed download happens in two stages. First the response headers come back, then the file bytes are read off the socket afterwards. The old `HttpClient.Timeout = 10 minutes` covered both stages, so a download that stalled mid-file died after 10 minutes. This PR changes that balance and only the first stage is now protected.

Here is the chain:

- `backend/CPS.ComplexCases.Egress/Extensions/IServiceCollectionExtension.cs:65` sets the storage client to `Timeout.InfiniteTimeSpan`, so the client itself no longer caps anything.
- The per-request `CancellationTokenSource` in `backend/CPS.ComplexCases.Egress/Client/BaseEgressClient.cs` only wraps the header read. On a streamed response `SendAsync` returns the moment the headers arrive, and the token source is disposed right after. So once you hold the stream, that token is already gone and the body read falls through to whatever the caller does.
- The caller does pass a token: `sourceStream.ReadAsync(partData..., cancellationToken)` at `backend/CPS.ComplexCases.FileTransfer.API/Durable/Activity/TransferFile.cs:293`. But that `cancellationToken` is the activity token from the `[ActivityTrigger]` method (`backend/CPS.ComplexCases.FileTransfer.API/Durable/Activity/TransferFile.cs:42-43`), and the only deadline behind it is `functionTimeout: "12:00:00"` in `backend/CPS.ComplexCases.FileTransfer.API/host.json:3`.

The practical effect: `SocketsHttpHandler` puts no read timeout on the response stream by default. Picture Egress accepting the connection, sending headers, then the network drops silently so no more bytes arrive and the socket is never closed. `ReadAsync` has no reason to return and the activity token is not signalled, so the transfer sits there for up to 12 hours before the function timeout finally cancels it. Before this change the same stall ended after 10 minutes.

Could you give the read a real deadline rather than relying on the activity token? Two ways:

- Wrap each `ReadAsync` in a `CancellationTokenSource(transferTimeout)` linked to the activity token, so a read that makes no progress fails in minutes.
- Or read the download through a stream that enforces its own idle timeout.

backend/CPS.ComplexCases.Egress/Models/EgressOptions.cs
```diff
@@ -6,4 +6,12 @@ public class EgressOptions
   public required string Username { get; set; }
   public required string Password { get; set; }
   public required string Url { get; set; }
+
+  // Management calls (workspace/list/permission/token)
+  public int ManagementTimeoutSeconds { get; set; } = 100;
+
+  // Applies to a single chunk upload (8-10 MB). Normally seconds; 5 minutes absorbs slow links
+  // without re-introducing the old 10-minute hang. Streamed downloads are not capped here -
+  // their body read is governed by the orchestration CancellationToken.
```
HasanCPS: This comment says streamed downloads are "governed by the orchestration CancellationToken," and the matching note at `BaseEgressClient.cs:57-60` says the token "only covers the header read." Both read like a guarantee, but the real deadline behind that token is the 12-hour `functionTimeout`, not a transfer-sized limit. Worth softening the wording so the next person is not misled (see the [download-timeout comment](https://github.com/CPS-Innovation/Large-and-Complex-Cases/pull/479/changes#r3443343412) for the full trace).

backend/CPS.ComplexCases.FileTransfer.API/Durable/Activity/TransferFile.cs
HasanCPS, line 108: The idle-timeout wrapper is a clean fix and the multipart path is well covered. One gap on the single-PUT path: the AWS SDK reads the wrapped stream, so a `TimeoutException` is thrown from inside the read while the SDK is consuming it. If the SDK rewraps stream exceptions, it will skip the `catch (TimeoutException)` and land in the general `catch (Exception)` as `GeneralError` rather than `Transient`, so it may not get retried. The stall itself still unblocks in minutes either way, so this is about error classification, not the hang. Could we add a single-PUT stall test, or confirm the SDK lets `TimeoutException` through unwrapped?

HasanCPS (approved):
LGTM!

PR #481, opened 2026-06-23
common resilence to stop object disposed exception

bridgesr:


backend/CPS.ComplexCases.Common/Resilience/SharedResiliencePolicy.cs
```diff
@@ -0,0 +1,36 @@
...
+  private readonly Func<ILoggerFactory, IAsyncPolicy<HttpResponseMessage>> _policyFactory;
+  private readonly object _gate = new();
+  private IAsyncPolicy<HttpResponseMessage>? _policy;
+
+  public SharedResiliencePolicy(Func<ILoggerFactory, IAsyncPolicy<HttpResponseMessage>> policyFactory)
+  {
+    _policyFactory = policyFactory ?? throw new ArgumentNullException(nameof(policyFactory));
+  }
+
+  public IAsyncPolicy<HttpResponseMessage> GetPolicy(IServiceProvider serviceProvider)
```
HasanCPS, line 23: ###  Double-checked lock on a non-volatile field 

`GetPolicy` reads `_policy` on the fast path outside the lock with no memory barrier:

```csharp
if (_policy is not null)
{
    return _policy;
}
```

`_policy` is not `volatile`. On a weak memory model a thread can observe a non-null reference before the writes that constructed the Polly policy are visible, returning a partially constructed object. This is benign on x64, which is almost certainly the deploy target, so it is a hardening nit rather than a live bug. If you want it airtight, mark `_policy` `volatile` or use `LazyInitializer.EnsureInitialized`. Not a blocker.

HasanCPS (approved):
LGTM!

PR #484, opened 2026-06-30
improve large transfer logic

bridgesr:


backend/CPS.ComplexCases.FileTransfer.API/Durable/Orchestration/TransferOrchestrator.cs
```diff
@@ -122,6 +123,31 @@ await context.Entities.CallEntityAsync(
...
+            {
+                var destinationFolderPaths = cleanFiles
+                    .Select(sp => EgressStorageClient.GetDestinationFolderPath(input.DestinationPath, sp.RelativePath, input.SourceRootFolderPath))
+                    .Where(p => !string.IsNullOrEmpty(p))
+                    .Distinct(StringComparer.OrdinalIgnoreCase)
+                    .ToList();
+
+                if (destinationFolderPaths.Count > 0)
+                {
+                    await context.CallActivityAsync(
```
HasanCPS: ## Pre-create folder activity has no retry and sits inside the fatal try/catch

The pre-create call goes out with no retry policy:

```csharp
await context.CallActivityAsync(
    nameof(CreateEgressDestinationFolders),
    new CreateEgressFoldersPayload { ... });
```

Inside the activity, `CreateFolderStructureAsync` calls `TryCreateFolderAsync`, which swallows a 409 but rethrows everything else, including a 500 or a `ManagementTimeoutSeconds` timeout (`EgressStorageClient.cs:380-388`). That exception has nowhere to go but the orchestrator's outer catch at `TransferOrchestrator.cs:322`, which marks the whole transfer `Failed` and rethrows.

So a single Egress blip on one folder segment during pre-creation kills the entire run with zero files transferred. That is the same "every file failed" outcome the ticket set out to remove, just triggered from a new spot. The old lazy path failed only the one file that hit the error, and it failed it as transient so the orchestrator could retry.

Fix: attach a `TaskOptions` retry policy to this activity call, or catch inside the activity and let files fall back to the lazy 404-recovery path that still exists in `InitiateUploadAsync`.

backend/CPS.ComplexCases.FileTransfer.API/Durable/Activity/TransferFile.cs
```diff
@@ -97,6 +97,27 @@ public async Task<TransferResult> Run([ActivityTrigger] TransferFilePayload payl
 
             using (sourceStream)
             {
+                if (totalSize <= 0)
```
HasanCPS, line 100: ## Zero-byte source files now fail fast as a general error 

```csharp
if (totalSize <= 0)
{
    ...
    return CreateFailureResult(..., TransferErrorCode.GeneralError, ...);
}
```

This is aimed at the `802ebfb5` production mode where the source reports 0 or unknown length, which is reasonable. The problem is it also rejects a genuinely empty 0-byte file. Before this change those went through the normal path. Now an empty file fails with a user message that says "please try again, and contact support if the problem continues," which a retry will never fix because the file really is empty.

If empty files can legitimately appear in an evidence set, split the two cases: fail on unknown length, allow a real 0-byte file. If empty files are out of scope, worth a one-line confirmation in the PR so it is a deliberate call rather than a side effect.

backend/CPS.ComplexCases.FileTransfer.API/Durable/Activity/TransferFile.cs
```diff
@@ -500,6 +528,20 @@ private TransferResult CreateFailureResult(
         return new TransferResult { IsSuccess = false, FailedItem = failedItem };
     }
 
+    // Maps an internal error code to a short, user-facing message. Detailed exception information is
+    // retained in telemetry and logs; only this concise message is surfaced to the user.
+    private static string MapUserMessage(TransferErrorCode errorCode) => errorCode switch
```
HasanCPS, line 537: `MapUserMessage` is applied to the two `GeneralError` paths (the generic catch at `:258` and the zero-byte guard), which is good. But the transient catches still push the raw string into the user-facing `ErrorMessage`:

```csharp
var errorMessage = $"Transient destination error (HTTP {(int)ex.StatusCode}): {ex.Message}";
```

and the S3 one, `"Transient S3 error (HTTP ..., ErrorCode=...)"`. A file that exhausts its retries and finally fails as transient still surfaces that raw HTTP-and-exception text to the user, which is the exact thing the ticket flagged as unusable. As a side effect, the `Transient`, `FileExists`, and `IntegrityVerificationFailed` arms of `MapUserMessage` are currently unreachable.

Route the final user-facing message through `MapUserMessage` for the transient paths too, and keep the raw detail in telemetry where it already goes.

backend/CPS.ComplexCases.FileTransfer.API/Durable/Orchestration/TransferOrchestrator.cs
```diff
@@ -122,6 +123,50 @@ await context.Entities.CallEntityAsync(
...
+                            nameof(CreateEgressDestinationFolders),
+                            new CreateEgressFoldersPayload
+                            {
+                                WorkspaceId = input.WorkspaceId!,
+                                FolderPaths = destinationFolderPaths,
+                                CaseId = input.CaseId,
+                                UserName = input.UserName,
+                                CorrelationId = input.CorrelationId
+                            },
+                            new TaskOptions(TaskRetryOptions.FromRetryPolicy(new RetryPolicy(
```
HasanCPS, line 155: This retry policy is the only hardcoded one in the change:

```csharp
new RetryPolicy(
    maxNumberOfAttempts: 3,
    firstRetryInterval: TimeSpan.FromSeconds(5),
    backoffCoefficient: 2.0)
```

Every other retry and delay you added is configurable (`SizeConfig.MaxOrchestratorRetries` and `RetryBatchSize`, `EgressOptions.MaxChunkUploadAttempts`, `ChunkRetryBaseDelaySeconds`, `CreateUploadRetryAttempts`, `CreateUploadSettleDelaySeconds`, `NetAppOptions.CredentialPropagationDelaySeconds`), so this is the one value we would need a redeploy to tune. The orchestrator already reads `_sizeConfig`, so could you pull the attempt count into a `SizeConfig` field (say `FolderPreCreateRetryAttempts`, default 3), and ideally the interval and backoff too? Keeps the tuning consistent with the rest of the ticket.

backend/CPS.ComplexCases.FileTransfer.API/Durable/Activity/TransferFile.cs
```diff
@@ -500,6 +529,20 @@ private TransferResult CreateFailureResult(
...
+    {
+        TransferErrorCode.FileExists =>
+            "A file with the same name already exists at the destination.",
+        TransferErrorCode.IntegrityVerificationFailed =>
+            "The file was uploaded but failed integrity verification, so the transfer was not completed.",
+        TransferErrorCode.Transient =>
```
HasanCPS, line 543:
```csharp
TransferErrorCode.Transient =>
    "The destination service was temporarily unavailable. The transfer will be retried automatically.",
```

A file only lands in the user's failed list after the orchestrator has used up its retries, so by the time someone reads this, nothing is going to retry it. The wording reads as "wait and it will sort itself out" when it will not. Suggest dropping the retry promise for this terminal case, something like:

```
The destination service was temporarily unavailable, so the file was not transferred. Please try again.
```

backend/CPS.ComplexCases.FileTransfer.API/Durable/Orchestration/TransferOrchestrator.cs
```diff
@@ -212,28 +259,40 @@ await context.Entities.CallEntityAsync(
...
-                            SourceRootFolderPath = input.SourceRootFolderPath,
-                            BearerToken = input.BearerToken,
-                            BucketName = input.BucketName,
-                            UserName = input.UserName!,
-                            CorrelationId = input.CorrelationId!
-                        })).ToList();
-
-                var retryResults = await Task.WhenAll(retryBatch);
-                await ProcessTransferResults(context, entityId, retryResults, transferOrchestrationEvent, isRetry: true);
+                // Re-dispatch as fresh TransferFile activities, but at reduced concurrency
```
HasanCPS, line 262: The comment here is honest about it: a whole-file retry re-initiates a fresh Egress upload session and leaves the previous partial `uploadId` behind, because there is no abort-upload API on the client. This is exactly the thing ticket action point 4 asked us to confirm ("confirm a re-initiated upload does not orphan the previous partial uploadId on Egress"), and the answer is that it does orphan it. On a 189-file batch that hits the retry path, Egress accumulates orphaned partial multipart sessions.

Not a crash and not a blocker, but worth deciding on explicitly rather than by default:

- If Egress expires abandoned sessions on its own (worth confirming), a short note in the comment saying so would close the action point.
- If it does not, consider adding an abort-upload call to the client and cancelling the previous `uploadId` before re-initiating, or reusing the existing session on retry instead of starting a fresh one.

At minimum, please confirm the Egress-side expiry behaviour so we know this is bounded.

backend/CPS.ComplexCases.FileTransfer.API/Durable/Activity/TransferFile.cs
```diff
@@ -174,19 +196,24 @@ public async Task<TransferResult> Run([ActivityTrigger] TransferFilePayload payl
             return CreateFailureResult(
                 payload.SourcePath.FullFilePath ?? payload.SourcePath.Path,
                 TransferErrorCode.Transient,
-                errorMessage,
+                MapUserMessage(TransferErrorCode.Transient),
                 ex);
         }
-        catch (HttpRequestException ex) when ((int?)ex.StatusCode >= 500)
+        catch (HttpRequestException ex) when ((int?)ex.StatusCode >= 500
```
HasanCPS, line 202: The 404 and 409 to transient reclassification is scoped by exception type, not by direction or operation. The `catch` fires for any `HttpRequestException` whose status is 404 or 409, anywhere in the transfer:

```csharp
catch (HttpRequestException ex) when ((int?)ex.StatusCode >= 500
    || ex.StatusCode == System.Net.HttpStatusCode.NotFound
    || ex.StatusCode == System.Net.HttpStatusCode.Conflict)
```

The intent is the Egress create-upload path, where a 404 means the destination folder is not visible yet and a 409 is a concurrent-create race. That part is correct. The gap is the other direction. On an `EgressToNetApp` transfer the Egress client is the source, so a genuine source 404 (a file listed and then deleted before the read) now gets classified as transient, retried by the orchestrator across the 60s, 120s, 240s backoff, and finally surfaced to the user as "temporarily unavailable." The file is gone, so about seven minutes of retries buy nothing and the message points at the wrong cause.

It is low severity. The `AmazonS3Exception` arm at `TransferFile.cs:189` already treats a source 404 as transient, so this is a consistent extension of existing behaviour rather than a fresh mistake, and the `EgressToNetApp` delete-during-read window is narrow. If you want to tighten it, gate the 404 and 409 branch on the destination being Egress, for example `destinationClient is EgressStorageClient`, so a source-side 404 keeps falling through to the general handler and fails fast with an accurate message.

HasanCPS (approved):
LGTM!

PR #485, opened 2026-07-01
Task/fct2 20209 test move functionality

lilachdavis:


e2e/pw/tests/egress-to-netapp-move-large-default.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+      await test.step(
+        `Verify file '${file.fileName}' is no longer present in Egress`,
+        async () => {
+          const exists = await isFileInEgress(
+            testData.egressBaseUrl,
+            testData.egressToken,
+            testData.workspace.id,
+            testData.destinationSubfolderId!,
+            file.fileName
+          );
```
HasanCPS, line 99: ## Step 10 checks the wrong Egress folder, so it can never fail

Step 10 is meant to prove the Move actually removed the files from Egress, but it
looks in the wrong folder.

The files are uploaded to and moved out of `4. Served Evidence/<uploadSubfolder>/`.
The check passes `testData.destinationSubfolderId`, which is the id for
`2. Counsel only/<uploadSubfolder>/` (see `setup-helper-default.ts:89`, where it's
captured for the NetApp → Egress *copy* teardown). The moved files were never in
that folder, so `listEgressWorkspaceFilesByFolderId` comes back empty and `exists`
is always `false`. The assertion passes whether or not the Move removed anything.

The source folder id isn't available today either: `setup-helper-default.ts:78`
calls `createFolder` for the source but throws the return value away. To make this
check real, capture the source folder id the same way the destination one is
captured, expose it on `TestSetupResult`, and pass that here instead.

```ts
// what it does now — checks 2. Counsel only, where the files never were
const exists = await isFileInEgress(
  testData.egressBaseUrl,
  testData.egressToken,
  testData.workspace.id,
  testData.destinationSubfolderId!,   // wrong folder
  file.fileName
);
```

Until this points at the source folder, Step 10 is a no-op.

.gitignore
```diff
@@ -334,14 +334,15 @@ nuget.config
 # postman
 *.postman_environment.json
 
-# E2E Postman Tests
-e2e/*/secrets.config.ps1
+# E2E Tests
+e2e/*/*.secrets.config.ps1
```
HasanCPS, line 338: ## `.gitignore` change stops ignoring the real pw secrets file

The pattern went from `e2e/*/secrets.config.ps1` to `e2e/*/*.secrets.config.ps1`.
The tracked template is `e2e/postman/secrets.config.template.ps1`, so the real file
is named `secrets.config.ps1` with no prefix. The new glob needs a prefix before
`.secrets.config.ps1`, so it no longer matches that name.

`git check-ignore e2e/pw/secrets.config.ps1` now returns nothing. Only the postman
copy is still covered, via the explicit `e2e/postman/secrets.config.ps1` line that
was added. A `secrets.config.ps1` dropped into `e2e/pw/` would be committable, which
is how credentials leak.

Either keep the old `e2e/*/secrets.config.ps1` line or add the pw path directly
alongside the postman one.

e2e/pw/pages/TransferMaterialsTab.ts
```diff
@@ -232,12 +232,14 @@ export class TransferMaterialsTab {
     while (Date.now() - start < timeout) {
       console.log(`  Waiting for ${fileName} to be indexed; refreshing...`);
       await this.page.waitForTimeout(5_000);
-      await this.page.reload();
+      // await this.page.reload();
```
HasanCPS: ##  Leftover commented-out `page.reload()`

The disabled `// await this.page.reload();`, the mis-indented comment under it, and
the extra blank line are dead code. The reason it was removed already lives in the
commit message. A single plain comment saying why reload is skipped reads better
than keeping the old call around commented out.

.gitignore
lilachdavis: Fixed

e2e/pw/pages/TransferMaterialsTab.ts
lilachdavis: Replaced with just a comment

e2e/pw/tests/egress-to-netapp-move-large-default.spec.ts
lilachdavis: Fixes in next commit:
- Capture and return `sourceSubfolderId` in the same manner as `destinationSubfolderId`
- Use `sourceFolderId` as the input for the function call `await isFileInEgress(params)`

PR #489, opened 2026-07-06
Story/fct2 18778 transfer material v1 UI integration test

renjithabby:
# PR checklist

## Correctness
- [ x] Does what the ticket asks for
- [x ] Handles the edge cases (empty, null, zero, large inputs), not just the happy path
- [x ] New logic has tests, including the failure cases
- [x ] Tests assert something real, not just run the code

## Keeping it simple
- [ x] Doesn't rebuild something that already exists in the codebase
- [ x] No more complex or slower than it needs to be
- [x ] No leftover debug logging or commented-out code

- [x ] One logical change, not a pile of unrelated stuff

### If you touched the UI

- [x ] New config or feature flags are wired into the VITE build variables, not just added in code
- [ x] Data fetches handle the loading, empty and error states, not just success, and match how the rest of the app already does it
- [x ] Ran prettier (npm run prettier:format), since CI lints but doesn't check formatting


ui-spa/src/common/hooks/useUserGroupsFeatureFlag.ts
```diff
@@ -92,5 +94,17 @@ export const useUserGroupsFeatureFlag = (): FeatureFlagData => {
     }),
     [groups, searchParams, userDetails.username],
   );
-  return useMemo(() => getFeatureFlags(), [getFeatureFlags]);
+
+  if (!featureFlags) {
```
HasanCPS: ##  Feature flags freeze after the first render 

The old hook ended with `return useMemo(() => getFeatureFlags(), [getFeatureFlags])`, so flags recomputed whenever `groups`, `searchParams`, or `username` changed. The new hook computes flags once inside `if (!featureFlags)`, dispatches them, and returns that value. Every later call reads the frozen copy from state.

That breaks two cases:

- MSAL account groups and `userDetails.username` populate asynchronously. If the first render captures them before they've loaded, the flags are computed from incomplete inputs and never recompute. An entitled user whose groups arrive a beat late never sees the feature switch on.
- The URL params (`?transfer-materials-v1`, `?case-details`, and the rest) feed into flag calculation. Client-side navigation that changes those params no longer re-evaluates the flags, so toggling a flag by URL mid-session does nothing, and once it's on it can't be turned off within the session.

The old `useMemo` existed precisely to react to these inputs. Freezing on first render throws that away.

ui-spa/src/common/hooks/useUserGroupsFeatureFlag.ts
```diff
@@ -92,5 +94,17 @@ export const useUserGroupsFeatureFlag = (): FeatureFlagData => {
...
+    const featureFlagsData = getFeatureFlags();
+    dispatch({
```
HasanCPS: ## The hook dispatches into a parent provider during render 

The `dispatch` on line 100 runs in the render body, not inside an effect or event handler. `useUserGroupsFeatureFlag` is called from `RootLayout`, and the dispatch targets `MainStateProvider`, a different component. React treats that as an illegal render-phase update to another component.

StrictMode is on in `main.tsx`, so React logs `Cannot update a component (MainStateProvider) while rendering a different component (RootLayout)` and defers the update. The first commit renders every flag-gated consumer with `featureFlags = null`. Users get a brief flash of the wrong state on load: the v0 Transfer Materials tab where v1 should be, the disconnect button hidden when it should show, and so on, until the correct flags land on the next render.

The fix is to move the initial dispatch into a `useEffect` so it runs after commit rather than during render.

ui-spa/src/components/case-management/transfer-materials-v1/index.tsx
```diff
@@ -686,6 +693,8 @@ const TransferMaterialsV1Page: React.FC<TransferMaterialsV1PageProps> = ({
     );
   }, [
     transferId,
+    netAppData,
```
HasanCPS: ## A second poll loop can start on top of the first 

The status-polling effect now lists `netAppData` and `egressData` in its dependency array. Those come from react-query folder listings configured with `staleTime: 0` and `gcTime: 0`, so a refetch hands back a fresh object each time. New identity, effect re-runs.

During an active transfer, `transferId` is still set. If the user switches browser tabs and comes back, react-query refetches the egress and netApp folders on window focus. That produces new `netAppData`/`egressData` references, the effect fires again, and a second `pollTransferStatus` loop starts alongside the first. When the transfer finishes, `handleStatusResponse` runs from both loops, so navigation and `handleFileTransferClear` fire twice.

The effect only needs `transferId` to decide when to start polling. The folder data shouldn't be a dependency here.

ui-spa/src/components/case-management/index.tsx
```diff
@@ -34,7 +34,8 @@ const CaseManagementPage = () => {
 
   const [activeTabId, setActiveTabId] = useState<TabId>("transfer-materials");
 
-  const featureFlags = useUserGroupsFeatureFlag();
+  const { state } = useContext(MainStateContext);
+  const { appData: { featureFlags } = {} } = state;
```
HasanCPS: ## Consumers read null flags on the first render 

`CaseManagementPage`, and the v1/v0 transfer pages, now read `featureFlags` from `MainStateContext` instead of calling the hook directly. Because of finding [here](https://github.com/CPS-Innovation/Large-and-Complex-Cases/pull/489/changes#r3528551134), that context is `null` until Layout's render-phase dispatch propagates on a later render.

On the first pass `featureFlags?.transferMaterialsV1` is `undefined`, so the `tabItems` memo adds the v0 "Transfer materials" tab. When the flags populate, it re-renders with the v1 tab. The result is a brief wrong-tab flash on initial load. Low severity on its own, but it's the same root cause as finding [here](https://github.com/CPS-Innovation/Large-and-Complex-Cases/pull/489/changes#r3528551134), and fixing the render-phase dispatch removes this flash too.

ui-spa/src/components/AppRoutes.tsx
```diff
@@ -16,9 +17,27 @@ import DisconnectSharedDriveConfirmationPage from "./case-management/netapp-disc
...
+    //dont want to update if featureFlags is already set as it need to be set only once
+    if (featureFlagsData && !featureFlags) {
+      dispatch({
+        type: "SET_FEATURE_FLAGS",
+        payload: {
+          featureFlags: featureFlagsData,
+        },
+      });
+    }
+  }, [featureFlagsData, featureFlags, dispatch]);
```
HasanCPS, line 39: ## Feature flags never update after the first render 

The effect dispatches the flags into state exactly once, then locks itself out:

```ts
useEffect(() => {
  //dont want to update if featureFlags is already set as it need to be set only once
  if (featureFlagsData && !featureFlags) {
    dispatch({ type: "SET_FEATURE_FLAGS", payload: { featureFlags: featureFlagsData } });
  }
}, [featureFlagsData, featureFlags, dispatch]);
```

`useUserGroupsFeatureFlag()` is reactive. It recomputes when the user's groups, the URL search params, or the username change. But once `featureFlags` is set in state, the `!featureFlags` guard is false from then on, so the effect drops every later value. Context keeps the first result for the rest of the session.

Two ways that bites:

- MSAL groups and username populate asynchronously. If the first render captures them before they've loaded, the flags compute from partial inputs and never correct themselves. An entitled user whose groups land a beat late doesn't get the feature.
- The flags read URL params (`?transfer-materials-v1` and the rest). Changing those by client-side navigation no longer re-evaluates anything, so toggling a flag mid-session does nothing until a full page reload.

That second point lands directly on this branch. If the integration test flips `?transfer-materials-v1=true` by navigating rather than reloading, it reads the old value and the v0 panel stays put. A hard reload with the param set works, which makes this read as flaky rather than broken.

**Suggested Fix:** drop the set-once guard so the effect re-dispatches whenever `featureFlagsData` changes. The cleaner version is under the shared root cause below.

ui-spa/src/components/Layout.tsx
```diff
@@ -29,12 +31,12 @@ export default function RootLayout({
...
     <div ref={skipLinkSiblingRef} tabIndex={-1} className={styles.rootLayout}>
       <SkipLink href="#main-content">Skip to main content</SkipLink>
-      {featureFlags.globalNav && (
+      {featureFlags?.globalNav && (
         <div className="govuk-width-container">
           <cps-global-header></cps-global-header>
         </div>
       )}
-      {!featureFlags.globalNav && <Header />}
+      {featureFlags !== null && !featureFlags?.globalNav && <Header />}
```
HasanCPS: ## One frame with no header on every page load 

The dispatch now happens in a post-paint effect, so `featureFlags` is `null` on the first render. Both header branches fall through on null:

```ts
{featureFlags?.globalNav && ( ...global header... )}              // null?.globalNav -> skip
{featureFlags !== null && !featureFlags?.globalNav && <Header />}  // featureFlags !== null is false -> skip
```

So the first commit paints no header at all. The effect then fires, `featureFlags` populates, and the header appears on the next render. That's a header pop-in on every load and refresh.

It's a regression from this refactor. When Layout read the flags straight from the hook, they were never null at paint, so there was no gap. The `featureFlags !== null` guard on line 39 exists to avoid flashing the wrong header during the null window, but the trade it makes is flashing no header instead.

**Suggested Fix:** give the reducer's initial `featureFlags` a computed default, or render the standard `Header` while `featureFlags` is null rather than gating on it. Same underlying cause as [finding 1](https://github.com/CPS-Innovation/Large-and-Complex-Cases/pull/489/changes#r3535282633).

ui-spa/src/components/AppRoutes.tsx
```diff
@@ -16,9 +17,27 @@ import DisconnectSharedDriveConfirmationPage from "./case-management/netapp-disc
...
+import { MainStateContext } from "../providers/MainStateProvider";
+import { useUserGroupsFeatureFlag } from "../common/hooks/useUserGroupsFeatureFlag";
 
 import ProtectedRoutes from "./ProtectedRoutes";
+
 const AppRoutes = () => {
+  const { state, dispatch } = useContext(MainStateContext);
+  const { appData: { featureFlags } = {} } = state;
+  const featureFlagsData = useUserGroupsFeatureFlag();
+  useEffect(() => {
```
HasanCPS, line 29: ## Shared root cause (worth a design call)

Findings [1](https://github.com/CPS-Innovation/Large-and-Complex-Cases/pull/489/changes#r3535282633) and [2](https://github.com/CPS-Innovation/Large-and-Complex-Cases/pull/489/changes#r3535321507) are the same decision seen twice. A value that a pure hook already derives gets copied into reducer state through an effect, then read back from context. That copy is what creates both the set-once staleness and the null-first-render gap, and it's why every consumer has to write `featureFlags?.` everywhere.

If the flags were derived where they're read, either computed inside `MainStateProvider` or handed out through a small `FeatureFlagContext`, consumers would read a live value. No effect, no set-once guard, no null window. That clears both findings at once rather than patching each symptom. It's a larger change, so call it as you see it, but it's the clean shape of what this refactor is reaching for. If you'd rather keep the change small, the two per-finding fixes above are enough on their own.

ui-spa/src/mocks/handlers.ts
```diff
@@ -92,15 +92,17 @@ export const setupHandlers = (baseUrl: string, apiMockSource: string) => {
...
+        const egressRootFolderResults = isDevMock()
           ? getEgressFolderResultsDev(folderId as string, path as string)
           : getEgressFolderResultsPlaywright(
               folderId as string,
               path as string,
             );
         await delay(500);
-
-        return HttpResponse.json(netAppRootFolderResults);
+        console.log("egressRootFolderResults:", egressRootFolderResults);
```
HasanCPS: ##  Debug console.logs left in the MSW handler (cleanup)

The egress-folder mock handler picked up three log lines:

```ts
console.log("folderId:", folderId);
console.log("path:", path);
// ...
console.log("egressRootFolderResults:", egressRootFolderResults);
```

The mock server runs in dev and Playwright builds, so every egress folder request dumps the id, the path, and the whole result payload to the console. Pull these before merge.

One thing in the same hunk is worth keeping: the variable was renamed from `netAppRootFolderResults` to `egressRootFolderResults`, which is what this handler actually returns. That rename is correct, leave it in.

ui-spa/package.json
HasanCPS: ## Unused dependency: use-reducer-async (cleanup)

`asyncActionHandlers.ts` and its only consumer (`MainStateProvider` via `useReducerAsync`) were both deleted in this branch. Nothing in `src` imports `use-reducer-async` anymore, so the dependency is dead weight in the lockfile now. Drop it so it doesn't leave the next person guessing which state mechanism is live.

ui-spa/src/common/hooks/useFormattedAreaValues.ts
```diff
@@ -14,11 +14,8 @@ const mapOption = (item: CaseDivisionsOrArea) => ({
 });
 
 export const useFormattedAreaValues = () => {
-  const {
-    state: {
-      apiData: { caseDivisionsOrAreas },
-    },
-  } = useMainStateContext()!;
+  const { state } = useContext(MainStateContext) ?? {};
+  const { apiData: { caseDivisionsOrAreas } = {} } = state;
```
HasanCPS: ## Dead `?? {}` fallback that would crash anyway (cleanup)

```ts
const { state } = useContext(MainStateContext) ?? {};
const { apiData: { caseDivisionsOrAreas } = {} } = state;
```

`MainStateContext` now ships a real default (`{ state: initialState, dispatch }`), so `useContext` never returns null and the `?? {}` branch can't be taken. And if it ever were, the next line destructures `state`, which would be `undefined`, and throws `Cannot destructure property 'apiData' of undefined`. So the guard is both dead and misleading: it implies a safe fallback that would actually crash on the following line. Drop the `?? {}` and read `state` directly.

ui-spa/src/components/case-management/transfer-materials-v1/index.tsx
```diff
@@ -672,19 +679,25 @@ const TransferMaterialsV1Page: React.FC<TransferMaterialsV1PageProps> = ({
...
     }
 
     setTransferStatus("transferring");
     pollTransferStatus(
-      transferId,
+      transferId!,
       isComponentUnmounted,
       handleStatusResponse,
       setApiRequestError,
     );
```
HasanCPS, line 701: ## Active transfer shows a blank panel until both folder listings come back 

When you open a case that already has an active transfer, the Transfer materials panel stays empty for as long as the two folder calls take to return. Here's the chain that gets you there.

`transferId` is seeded from `activeTransferId` on the first render (line 144), so it's truthy right away. With a transfer active:

- the source view is skipped, because it's gated on `!transferId` (line 751)
- `transferStatus` is still `null`, so `renderActiveTransferMessage` returns an empty div. The spinner in there only shows for `"transferring"` or `"validating"` (line 573 / 586)

The thing that would fill that gap is the `"transferring"` status, and it's now set inside the effect that waits on `shouldStartTransferStatusPolling`:

```ts
const shouldStartTransferStatusPolling = useMemo(
  () => Boolean(transferId && netAppData && egressData),   // line 682
  [transferId, netAppData, egressData],
);

useEffect(() => {
  if (!shouldStartTransferStatusPolling) return;
  setTransferStatus("transferring");                       // line 692
  pollTransferStatus(transferId!, isComponentUnmounted, handleStatusResponse, setApiRequestError);
}, [shouldStartTransferStatusPolling, transferId, setTransferStatus, isComponentUnmounted, handleStatusResponse]);
```

`netAppData` and `egressData` are the folder listings from `getNetAppFolders` / `getEgressFolders` (lines 166, 208). Both queries run with `staleTime: 0` and `gcTime: 0`, so nothing is cached and they refetch from the network on every mount. Until both responses land, the guard is false, the effect bails, and the panel is blank. The case chrome and tab bar are still there, so it's the panel that's empty rather than the whole screen.

On `main` this didn't happen. The effect read `if (!transferId) return; setTransferStatus("transferring");` with no data dependency, so the spinner showed the frame after paint. Now the wait is however long the two folder calls take, which on a slow connection is a visible blank.

One note on the line 683 comment (waiting so we don't show status to users without permission): that case is already handled. A permission failure comes back as a 401 on the folder query, and the error effect redirects those users to the connection-error page (lines 353 and 378). So the guard isn't shielding anyone here. All it does is delay the spinner for users whose folder calls are going to succeed anyway. And the transferring indicator doesn't need the folder contents at all, since those drive the source browser for picking files, not the progress view.

**Suggested fix** — set the status off `transferId`, and keep only the poll behind the folder guard:

```ts
useEffect(() => {
  if (!transferId) return;
  setTransferStatus("transferring");
}, [transferId, setTransferStatus]);

useEffect(() => {
  if (!shouldStartTransferStatusPolling) return;
  pollTransferStatus(transferId!, isComponentUnmounted, handleStatusResponse, setApiRequestError);
}, [shouldStartTransferStatusPolling, transferId, isComponentUnmounted, handleStatusResponse]);
```

The spinner comes back immediately, and the poll still keys off `shouldStartTransferStatusPolling` rather than the raw `netAppData` / `egressData`, so a focus refetch that changes their identity won't start a second poll loop. That's the behaviour the memoised boolean was added for, and this keeps it.

ui-spa/playwright/integration-tests/pages/transfer-material-source.ts
```diff
@@ -0,0 +1,314 @@
...
+          activityFileSections.nth(index).getByTestId("transfer-relative-path"),
+        ).toHaveText(relativePath.folderPath);
+        await expect(
+          activityFileSections.nth(index).locator("ul").locator("li"),
+        ).toHaveCount(relativePath.files.length);
+        const fileItems = activityFileSections
+          .nth(index)
+          .locator("ul")
+          .locator("li");
+        relativePath.files.forEach(async (file, i) => {
```
HasanCPS: ##  Success-banner file names aren't actually being checked 

`validateTransferSuccessBanner` (line 254) checks the file list under each folder in the success banner. Three of its four assertions are awaited fine: the section count (line 284), each section's relative-path label (287), and each section's file count (290). The fourth one, the check on the actual file names, is fire-and-forget:

```ts
relativePath.files.forEach(async (file, i) => {   // line 297
  await expect(fileItems.nth(i)).toHaveText(file);
});
```

`forEach` ignores whatever its callback returns. The callback is `async`, so each call produces a promise and `forEach` drops all of them. Walking it through:

1. `forEach` starts every callback, each fires its `expect(...).toHaveText(file)`, hands the promise back, and `forEach` discards it.
2. `forEach` returns `undefined` synchronously.
3. That's the last line of the `async (relativePath, index) => {…}` map callback, so the callback resolves the moment `forEach` returns. It never waits for the name checks.
4. `Promise.all` (line 285) awaits those map callbacks, which have already resolved, so `await Promise.all(...)` completes while the name checks are still pending in the background.
5. The test moves on to line 302.

So the per-file name assertions run detached from the test. If a name is wrong or the order is off, `toHaveText` does eventually reject, but Playwright polls up to the timeout and the rejection lands late, after the test has walked past. It surfaces as a floating unhandled rejection rather than failing this assertion, and on a lot of runs it's swallowed or pinned on the wrong test. On a passing run the promises settle quietly and disappear.

The result: the banner test confirms *how many* files sit under each folder, but not *which* files. Every caller passes an exact `files: string[]` expecting the names to be verified, and that isn't happening. Rename or reorder files in the banner and this test stays green.

**Suggested Fix** — await the inner checks by swapping the `forEach` for a `map` inside the existing `Promise.all`:

```ts
await Promise.all(
  relativePath.files.map((file, i) =>
    expect(fileItems.nth(i)).toHaveText(file),
  ),
);
```

A plain `for` loop with `await` works too. Either way the name checks block the test the way the count check already does.

HasanCPS (approved):
LGTM!

PR #490, opened 2026-07-07
story/FCT2-20389: Handle CaseMetaData operation name is null value

renjithabby:
https://cpsgovuk.atlassian.net/browse/FCT2-20389

# PR checklist

## Correctness
- [ x] Does what the ticket asks for
- [ x] Handles the edge cases (empty, null, zero, large inputs), not just the happy path
- [ x] New logic has tests, including the failure cases
- [ x] Tests assert something real, not just run the code

## Keeping it simple
- [ x] Doesn't rebuild something that already exists in the codebase
- [ x] No more complex or slower than it needs to be
- [x ] No leftover debug logging or commented-out code

- [ x] One logical change, not a pile of unrelated stuff

### If you touched the UI
-- [ x] Data fetches handle the loading, empty and error states, not just success, and match how the rest of the app already does it
- [x ] Ran prettier (npm run prettier:format), since CI lints but doesn't check formatting

### DIFF
--- a/ui-spa/playwright/integration-tests/transfer-materials-v1/transfer-material-egress-list.spec.ts
+++ b/ui-spa/playwright/integration-tests/transfer-materials-v1/transfer-material-egress-list.spec.ts
@@ -3,13 +3,10 @@ import { test } from "../utils/test";
 import { TransferMaterialsSourcePage } from "../pages/transfer-material-source";
 
 test.describe("transfer material egress list", () => {
-  test.beforeEach(async ({ page }) => {
-    await page.goto("/case/12/case-management?transfer-materials-v1=true");
-  });
-
   test("Should show the transfer material tab with correct initial content", async ({
     page,
   }) => {
+    await page.goto("/case/12/case-management?transfer-materials-v1=true");
     const transferMaterialsSourcePage = new TransferMaterialsSourcePage(page);
     await transferMaterialsSourcePage.verifyUrl("/case/12/case-management");
     await transferMaterialsSourcePage.verifyPageElements();
@@ -38,6 +35,7 @@ test.describe("transfer material egress list", () => {
   test("Should correctly navigate through the egress folders and validate checkbox visibility", async ({
     page,
   }) => {
+    await page.goto("/case/12/case-management?transfer-materials-v1=true");
     const transferMaterialsSourcePage = new TransferMaterialsSourcePage(page);
     await transferMaterialsSourcePage.verifyUrl("/case/12/case-management");
     await transferMaterialsSourcePage.verifyPageElements();
@@ -158,6 +156,7 @@ test.describe("transfer material egress list", () => {
     ]);
     await transferMaterialsSourcePage.verifyCheckboxesVisibility(false, 4);
   });
+
   test("Should show the leadDefendant name in the Home path for Egress if the operation name is null", async ({
     page,
     worker,
@@ -176,6 +175,7 @@ test.describe("transfer material egress list", () => {
         });
       }),
     );
+    await page.goto("/case/12/case-management?transfer-materials-v1=true");
     const transferMaterialsSourcePage = new TransferMaterialsSourcePage(page);
     await transferMaterialsSourcePage.verifyUrl("/case/12/case-management");
     await transferMaterialsSourcePage.verifyPageElements("John Doe");
--- a/ui-spa/playwright/integration-tests/transfer-materials-v1/transfer-material-egress-list.spec.ts
+++ b/ui-spa/playwright/integration-tests/transfer-materials-v1/transfer-material-egress-list.spec.ts
@@ -1,3 +1,4 @@
+import { delay, HttpResponse, http } from "msw";
 import { test } from "../utils/test";
 import { TransferMaterialsSourcePage } from "../pages/transfer-material-source";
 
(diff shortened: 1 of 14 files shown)


ui-spa/playwright/integration-tests/transfer-materials-v1/transfer-material-egress-list.spec.ts
```diff
@@ -157,4 +158,36 @@ test.describe("transfer material egress list", () => {
     ]);
     await transferMaterialsSourcePage.verifyCheckboxesVisibility(false, 4);
   });
+  test("Should show the leadDefendant name in the Home path for Egress if the operation name is null", async ({
```
HasanCPS, line 160: ## The egress Playwright test never applies its null-operationName override

The new test sets up a `worker.use` override that returns `operationName: null, leadDefendantName: "John Doe"`, then asserts the page shows `"John Doe"`. But the `describe` block has a `beforeEach` (line 6) that already calls `page.goto(...)`, and this test never navigates again after the override:

```tsx
test.beforeEach(async ({ page }) => {
  await page.goto("/case/12/case-management?transfer-materials-v1=true");
});
...
test("Should show the leadDefendant name ... if the operation name is null", async ({ page, worker }) => {
  await worker.use( /* operationName: null, leadDefendantName: "John Doe" */ );
  // no page.goto here
  await transferMaterialsSourcePage.verifyPageElements("John Doe");
});
```

MSW picks a handler when the request is intercepted. The metadata request fires during the `beforeEach` navigation, before the override is registered, so it is served by the default `"Thunderstruck"` handler. The test then registers the override but nothing refetches, so the assertion runs against whatever the first load returned. At best this is racy, at worst it passes against the wrong data and gives false confidence in the headline feature.

The sibling shared-drive test gets this right: `transfer-material-netapp-list.spec.ts` calls `await page.goto(...)` after `worker.use(...)`. Copy that ordering here so the override is in place before the page loads.

ui-spa/src/apis/gateway-api.test.ts
```diff
@@ -909,6 +910,30 @@ describe("gateway apis", () => {
         json: async () => mockData,
       });
 
+      await expect(getCaseMetaData("12")).rejects.toBeInstanceOf(ApiError);
+      await expect(getCaseMetaData("12")).rejects.toThrow(
+        "An error occurred contacting the server at gateway_url/api/v1/cases/12: response schema validation failed; status - OK (200)",
+      );
+    });
+    it("getCaseMetaData - response schema validation should fail if both operationName and leadDefendantName are null", async () => {
```
HasanCPS, line 918: ## The both-null schema unit test does not exercise the guard it claims to cover

The test says it checks that validation fails when `operationName` and `leadDefendantName` are both null, but the mock fails the base schema long before the both-null `superRefine` runs:

```tsx
const mockData = [          // an array; the schema is a single object
  {
    operationName: null,
    caseId: "123",          // string; schema expects number
    leadDefendantName: null,
    netappFolderPath: null, // null; schema field is non-nullable
    // activeTransferId missing
    registrationDate: null, // not in the schema
  },
];
```

Zod runs the object-level `superRefine` only after the shape parses. Passing an array fails immediately, and the string `caseId` and null `netappFolderPath` would fail the shape too. So the parse fails on type errors and the both-null message is never reached. The assertion only checks the generic `"response schema validation failed"` wrapper, which shows up for any invalid payload, so this test would still pass if the both-null guard were deleted.

Suggested fix: test the guard directly against a payload that is valid apart from the two null names, and assert on the specific issue:

```tsx
const result = caseMetaDataResponseSchema.safeParse({
  caseId: 12,
  egressWorkspaceId: "abcdef",
  netappFolderPath: "netapp/",
  operationName: null,
  leadDefendantName: null,
  activeTransferId: null,
  urn: "urn",
});
expect(result.success).toBe(false);
expect(result.error?.issues.some(i =>
  i.message.includes("cannot both be null"))).toBe(true);
```

That way removing or breaking the guard actually fails the test.

ui-spa/src/components/case-management/index.tsx
```diff
@@ -51,6 +51,9 @@ const CaseManagementPage = () => {
     staleTime: 0,
     gcTime: 0,
   });
+  const operationNameOrDefendantName = useMemo(() => {
+    return caseMetaData?.operationName || caseMetaData?.leadDefendantName || "";
```
HasanCPS, line 55: ## An empty-string operationName slips past the both-null guard and renders blank

With the schema at `ui-spa/src/schemas/responses/caseMetaData.ts:9`

The guard blocks the case where both names are `null`:

```ts
if (val.operationName === null && val.leadDefendantName === null) { ... }
```

The component coalesces with `||`, which is a falsy check, not a null check:

```tsx
return caseMetaData?.operationName || caseMetaData?.leadDefendantName || "";
```

So a response of `operationName: ""` with `leadDefendantName: null` passes validation (neither is `null`) but the runtime falls all the way through to `""`. The heading is blank and the `operation-name` param is empty, which is the empty-workspace-title state the guard was added to prevent.

Low likelihood since it needs the backend to send an empty string, but the guard and the coalescing disagree about what "missing" means. Either normalise empty to null on the way in, or have the guard treat empty string the same as null (for example check `!val.operationName?.trim() && !val.leadDefendantName?.trim()`), so the two stay consistent.

ui-spa/src/mocks/data/caseMetaData.playwright.ts
```diff
@@ -4,6 +4,7 @@ export const caseMetaDataPlaywright: CaseMetaDataResponse = {
   egressWorkspaceId: "egress_1",
   netappFolderPath: "netapp/",
   operationName: "Thunderstruck",
+  leadDefendantName: "null",
```
HasanCPS: ## Playwright fixture uses the string "null" for the defendant name

```ts
leadDefendantName: "null",
```

This is the literal text `"null"`, not an absent value and not a real name. It is dormant today because the fixture's `operationName` is `"Thunderstruck"`, so the fallback never fires. But if any flow reuses this fixture with a null or empty operation name, the heading and the `operation-name` param would render the word `null`. Set it to a real name (the dev fixture uses `"John Doe"`) or to actual `null` if that is what you meant to represent.

ui-spa/src/apis/gateway-api.test.ts
renjithabby: fixed.

ui-spa/playwright/integration-tests/transfer-materials-v1/transfer-material-egress-list.spec.ts
renjithabby: fixed

ui-spa/src/components/case-management/index.tsx
renjithabby: this is meant to be falsy check as null check is already covered by schema, and here what it is doing if an operationName is there show it or default to defendantLastName or default to "".

HasanCPS (approved):
LGTM!

PR #494, opened 2026-07-13
defect/FCT2-20551 Remove disconnect cancel btn

renjithabby:
https://cpsgovuk.atlassian.net/browse/FCT2-20551
# PR checklist

## Correctness
- [x ] Does what the ticket asks for
- [x ] Handles the edge cases (empty, null, zero, large inputs), not just the happy path
- [ x] New logic has tests, including the failure cases
- [x ] Tests assert something real, not just run the code

## Keeping it simple
- [x ] Doesn't rebuild something that already exists in the codebase
- [ ] xNo more complex or slower than it needs to be
- [ x] No leftover debug logging or commented-out code

- [ x] One logical change, not a pile of unrelated stuff

### If you touched the UI

- [ x] API changes are reflected in the MSW handlers in src/mocks and the matching zod schema, so the mocked tests aren't passing against a contract that's gone

- [ x] Ran prettier (npm run prettier:format), since CI lints but doesn't check formatting

### DIFF
--- a/ui-spa/playwright/integration-tests/disconnect-shared-drive.spec.ts
+++ b/ui-spa/playwright/integration-tests/disconnect-shared-drive.spec.ts
@@ -33,11 +33,7 @@ test.describe("disconnect-shared-drive", () => {
     await expect(page.locator("label").nth(1)).toHaveText(
       "No, keep this folder connected",
     );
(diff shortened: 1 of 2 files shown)


HasanCPS (approved):
LGTM!

PR #495, opened 2026-07-14
Story/FCT2- 20179 transfer DAS fixes

renjithabby:
# PR checklist

## Correctness
- [ X] Does what the ticket asks for
- [ X] Handles the edge cases (empty, null, zero, large inputs), not just the happy path
- [ X] Errors are handled, not swallowed
- [ X] New logic has tests, including the failure cases
- [X ] Tests assert something real, not just run the code

## Keeping it simple
- [ X] Doesn't rebuild something that already exists in the codebase
- [ X] No more complex or slower than it needs to be
- [ X] No leftover debug logging or commented-out code

- [ X] One logical change, not a pile of unrelated stuff

### If you touched the UI
- [X ] API changes are reflected in the MSW handlers in src/mocks and the matching zod schema, so the mocked tests aren't passing against a contract that's gone
- [ X] New config or feature flags are wired into the VITE build variables, not just added in code
- [ X] Data fetches handle the loading, empty and error states, not just success, and match how the rest of the app already does it
- [X ] Ran prettier (npm run prettier:format), since CI lints but doesn't check formatting
- [ X] Checked accessibility (keyboard nav, screen reader, sensible labels), since the Lighthouse run is manual and won't flag it for you


ui-spa/src/components/Layout.module.scss
```diff
@@ -1,20 +1,22 @@
...
-  max-width: 87.5rem;
-  padding-right: 1.5625rem;
-  padding-left: 1.5625rem;
-  margin: 0 auto;
 }
 .rootLayout {
+  & :global(.govuk-footer) {
+    margin-top: 1.875rem;
+  }
+  & :global(.govuk-button) {
```
HasanCPS: ##  The button width override lands on every page in the app

This rule is inside `.rootLayout`, and `Layout.tsx:35` puts `.rootLayout` on the app's outermost div in both flag states. That makes it an app-wide override of a GOV.UK Frontend rule rather than something scoped to the new layout.

GOV.UK Frontend makes buttons full width on phones on purpose, and only lets them shrink from tablet up. From `govuk-frontend/govuk/components/button/_index.scss`:

```scss
.govuk-button {
  width: 100%;
  @include govuk-media-query($from: tablet) {
    width: auto;
  }
}
```

The new rule compiles to `.Layout_rootLayout__hash .govuk-button`, which is specificity 0,2,0 against govuk's 0,1,0, and it carries no media query, so it wins at every width. Above tablet you will not see much, because `auto` and `fit-content` shrink-wrap an inline-block button about the same. Below tablet it strips the full-width behaviour from every button in the product, including the flag-off V0 journey this PR is otherwise careful not to change.

Worked example: open the case search page at 375px with `transferMaterialsV1` off. Before this branch the Search button spans the column. After it, the button is only as wide as the word "Search".

Worth noting `.controlWrapper` already carries the identical rule at `transfer-materials-v1/index.module.scss:63`, scoped to the place that actually needs it.

Could we drop the root-level copy and let the component-level rule do the work, so mobile buttons elsewhere keep the standard GOV.UK sizing?

ui-spa/src/components/search-result-page/index.tsx
HasanCPS, line 378: ## `.btnWrapper` is gone from the stylesheet but both search pages still ask for it

The same pair exists at `ui-spa/src/components/egress-connect/EgressSearchPage.tsx:102`.

The `.btnWrapper { min-height: 4.375rem; }` rule came out of both modules, but both consumers still reference it. `styles.btnWrapper` is now `undefined`, so React drops the attribute and the div renders with no class at all. There is no `class="undefined"` in the DOM, just a wrapper that does nothing.

The part users will see is the lost `min-height`. It was reserving height on the button side of the row so the button lined up with the input. The same hunk also changed `align-items: flex-end` to `center` on `.inputWrapper` (`index.module.scss:9`, `EgressSearchPage.module.scss:4`), and those two changes work against each other.

Worked example: submit an empty search on the case search page. The govuk form group grows taller because the error message gets inserted above the input. `.inputWrapper` is the flex row holding `renderSearchForm()` and the button wrapper, so the row grows with it. With `flex-end` the button stayed down at the input's bottom edge. With `center` and no `min-height`, the button floats up to the middle of the taller row and no longer sits level with the input it belongs to.

Which way was this meant to go? If the reserved space is not wanted any more, could we delete the two dead wrapper divs as well? If it is still wanted, the rule needs to come back.

ui-spa/src/components/Layout.tsx
```diff
@@ -29,15 +29,21 @@ export default function RootLayout({
     }
   }, [location.pathname]);
   return (
-    <div ref={skipLinkSiblingRef} tabIndex={-1} className={styles.rootLayout}>
+    <div
+      ref={skipLinkSiblingRef}
+      tabIndex={-1}
+      className={`${styles.rootLayout} ${!featureFlags?.transferMaterialsV1 ? styles.rootLayoutV0 : ""}`}
```
HasanCPS: ## The V0 width class is applied before the feature flags load

`featureFlags` starts as `null` (`mainStateReducer.ts:16`) and only gets filled in by an effect after the first paint. `!featureFlags?.transferMaterialsV1` is `true` while it is still null, so `.rootLayoutV0` goes on during that window and comes off once the flags arrive.

For a user with the flag on, the page renders at the V0 width first (max-width 87.5rem) and then snaps back to the GOV.UK default once flags resolve. On a wide monitor that is a visible reflow on every page load, and anything the user was about to click shifts under the pointer.

The codebase already handles this for this exact flag, at `case-management/index.tsx:141`:

```tsx
if (featureFlags !== null && !featureFlags?.transferMaterialsV1) {
```

The `featureFlags !== null` guard is there because a negated flag check reads as true while loading. The other `featureFlags?.x` uses nearby are positive checks, so they do not have the problem and do not need the guard. This line is a negated check, so it does.

Can we add the same `featureFlags !== null` guard here?

ui-spa/src/components/Layout.module.scss
```diff
@@ -1,20 +1,22 @@
...
+    width: fit-content;
+  }
+}
+
+.rootLayoutV0 {
```
HasanCPS, line 13: ## `transferMaterialsV1` now changes page width on routes it has nothing to do with

The 87.5rem override used to apply unconditionally. Now it only applies when `transferMaterialsV1` is off, which quietly turns that flag into an app-wide page width switch.

`.govuk-width-container` is not just the transfer materials page. It also wraps the Header (`Header.tsx:8`), the Footer (`Footer.tsx:9`), case search results, egress connect, netapp connect, and the case management tabs. Turning the flag on for a user shrinks all of those from 1400px to the GOV.UK default, including pages that have nothing to do with transfer materials, and the header and footer will no longer line up with the content of any V0-era page.

This one may well be intended. The "gave extra width for the v0 layout" commit reads like the width change is meant to ride along with V1 everywhere, in which case ignore me. If the flag was only meant to change the transfer materials journey, could we scope the width rule to that route's own wrapper instead of the root layout?

ui-spa/src/components/case-management/transfer-materials-v1/index.module.scss
```diff
@@ -57,11 +57,20 @@
 }
 
 .controlWrapper {
```
HasanCPS, line 59: ## The disconnect button no longer has anything keeping it off the transfer controls

`.controlWrapper` used to be `display: flex; justify-content: space-between`, and that was what held TransferControls and the Disconnect button apart. The wrapper is a plain block now, and the button is taken out of flow above 50rem:

```scss
.disconnectButton {
  position: relative;
  margin: 0.65rem 0 0 0;
  @media (min-width: 50rem) {
    position: absolute;
    right: 0;
    bottom: 0;
  }
}
```

Out of flow means TransferControls no longer knows the button is there. TransferControls (`index.tsx:778`) is a left-aligned flex row of the direction toggle plus Copy and Move, and it now spans the full wrapper width. Nothing stops the two meeting in the middle, whereas flexbox used to guarantee they could not.

I have not proved a break, because whether they actually collide depends on text widths at a given viewport. The place to look is just above the 50rem breakpoint (an 800 to 900px tablet, or a narrowed desktop window) with `disconnectSharedDrive` on, where the wrapper is at its narrowest while the button is still absolutely positioned. If Copy and Move do reach that far they will sit under the Disconnect button, and a mis-aimed click disconnects the drive.

Could you check that range with the flag on? If there is any overlap, keeping `space-between` or reserving right padding on the controls would put the guarantee back.

ui-spa/src/components/netapp-connect/NetAppFolderResultsPage.module.scss
```diff
@@ -5,6 +5,9 @@
     text-align: right;
   }
   .tableContainer {
```
HasanCPS, line 7: ## One table wrapper got `max-width` without the `overflow-x` that makes it work

This is the only table wrapper in the PR that gets `max-width` on its own:

```scss
.tableContainer {
  max-width: 100%;
  @media (min-width: 40.0625em) {
    max-width: 66%;
  }
}
```

`NavigationTable.module.scss`, `SearchResults.module.scss` and `EgressSearchResults.module.scss` all pair the two so a wide table scrolls inside its own box. Without the pair here, a folder table with long names overflows `.tableContainer` and pushes the whole page into horizontal scroll instead, which is the thing the rest of the branch is fixing. The `max-width: 66%` above tablet makes that more likely on desktop rather than less, since the container is well narrower than the viewport.

Can we add `overflow-x: auto` here to match the other three?

ui-spa/src/components/Layout.module.scss
renjithabby: this is intentional,  we no longer need the extra width with v1.

ui-spa/src/components/case-management/transfer-materials-v1/index.module.scss
renjithabby: i have checked it wont  overlap it come closer and then disconnect move to down to the next row

ui-spa/src/components/netapp-connect/NetAppFolderResultsPage.module.scss
renjithabby: fixed

ui-spa/src/components/search-result-page/index.tsx
renjithabby: fixed

HasanCPS (approved):
LGTM!

PR #496, opened 2026-07-14
FCT2- 20332: Transfer material changes for new UI

kmshrajcps:
Adding  `TRANSFER_MATERIALS_V1` switch to the `e2e/pw` suite so it    
  drives the redesigned (v1) Transfer Materials screen when the flag is on and the  old screen when off — keeping the old screen covered during rollout and adding  
  new-screen
  coverage, without two suites fighting over selectors.

  - Screen-agnostic `TransferMaterialsTabApi` contract;
  `getTransferMaterialsTab`
    picks `TransferMaterialsTab` (old) or `TransferMaterialsTabV1`     
  (new) from the flag,
    so specs never reference a screen-specific selector.
  - New-screen page objects: `TransferMaterialsTabV1` +
  `TransferDestinationPage`
    (folder-tree confirm — no modal). Covers the "shared drive" table  
  ids,
    `Copy selected` / `Move selected`, the `View Shared Drive` / `View 
  Egress`
    toggle, and the new `transfer-errors` /
  `transfer-permissions-error` /
    `transfer-resolve-file-path` routes.
  - Recover from the shared full-page transfer error (both screens) so 
  the seed can
    re-enter the tab.
  - README: how the flag reaches the environment (SPA build flag +     
  private-beta
    group), corrected register-case file sizes, and the
  page-object-layer notes.

  ## Testing
  Ran the full `e2e/pw` suite locally against dev on both screens:     
  - **New UI:** existing-case 4/4, register-case 8/8
  - **Old UI:** existing-case 4/4, register-case 8/8

### DIFF
--- a/e2e/pw/README.md
+++ b/e2e/pw/README.md
@@ -214,15 +214,19 @@ Selects which screen the specs drive: `false` (default) = old screen, `true` =
 redesigned screen. It is a **selector only** — it must match what the
 environment renders for `E2E_AD_USER`, or the specs drive the wrong screen and fail.
 
-Specs never branch on this flag directly. They build the Transfer Materials
-page object via `getTransferMaterialsTab(page)`, which reads the switch and
-returns either `TransferMaterialsTab` (old screen) or `TransferMaterialsTabV1`
-(new screen). Both implement the shared `TransferMaterialsTabApi` contract, so
-a spec depends only on that surface and stays screen-agnostic. The new screen
-adds a destination-tree step after Copy/Move — driven by `TransferDestinationPage`
-— because it has no confirm modal. A spec that must diverge (e.g.
-`netapp-to-egress-copy-default`) reads `loadEnvConfig().transferMaterialsV1`
-explicitly rather than referencing screen-specific selectors.
+Specs stay screen-agnostic by building the Transfer Materials page object via
+`getTransferMaterialsTab(page)`, which reads the switch and returns either
+`TransferMaterialsTab` (old screen) or `TransferMaterialsTabV1` (new screen).
+Both implement the shared `TransferMaterialsTabApi` contract, so a spec depends
+only on that surface — no screen-specific selectors.
+
+The one place the flow genuinely differs is the NetApp → Egress destination:
+the old screen navigates the second Egress panel and confirms in a modal, while
+the new screen has no second panel and instead navigates a destination tree
+(`TransferDestinationPage`). The two specs that do this —
+`netapp-to-egress-copy.spec.ts` and `netapp-to-egress-copy-default.spec.ts` —
+branch on `loadEnvConfig().transferMaterialsV1` to pick the right path, rather
+than referencing screen-specific selectors directly.
 
 For a real Azure AD user the new screen renders only when **both** hold:
 
--- a/e2e/pw/README.md
+++ b/e2e/pw/README.md
@@ -13,20 +13,31 @@ Tests run in two modes:
 
 ### Test Matrix
 
-| Test | Files | Default Mode | Register Case Mode |
-|------|-------|--------------|--------------------|
-| Egress to NetApp Copy | 10MB x 1 | ✓ | ✓ |
-| Egress to NetApp Copy - Large | 50MB x 1 | ✓ | ✓ |
-| Egress to NetApp Copy - Multifile | 10MB x 3 | -- | ✓ |
-| Egress to NetApp Move | 10MB x 1 | ✓ | ✓ |
-| NetApp to Egress Copy | 10MB x 1 | ✓ (uses seeded fixture) | ✓ (sort + row 0) |
-| Full Flow (login, search, connect) | 10MB x 1 | -- | ✓ |
+The two modes use different file sizes for the same logical test — default
+mode stays small for fast iteration, register-case mode exercises larger
+transfers. The Default / Register columns show each mode's size (`--` = not
+run in that mode).
+
+| Test | Default Mode | Register Case Mode |
+|------|--------------|--------------------|
+| Egress to NetApp Copy | ✓ 10MB x 1 | ✓ 100MB x 1 |
+| Egress to NetApp Copy - Large | ✓ 50MB x 1 | ✓ 200MB x 1 |
+| Egress to NetApp Copy - Multifile | -- | ✓ 10MB x 3 |
+| Egress to NetApp Move | ✓ 10MB x 1 | ✓ 100MB x 1 |
+| NetApp to Egress Copy | ✓ 10MB x 1 (uses seeded fixture) | ✓ 100MB x 1 (sort + row 0) |
+| Full Flow (login, search, connect) | -- | ✓ 100MB x 1 |
+
+Default-mode sizes come from `TEST_FILE_SIZE_MB` (10) and
+`LARGE_TEST_FILE_SIZE_MB` (50); register-case sizes are set per spec via
+`test.use({ testOptions: { fileSizeMb, fileCount } })`.
 
 NetApp -> Egress Move is descoped — the product UI doesn't expose a Move
-button in that direction (`EgressFolderContainer.tsx` has no Move
-references; only `NetAppFolderContainer.tsx` does, gated on the
+button in that direction on either screen. Old screen: `EgressFolderContainer.tsx`
+has no Move references; only `NetAppFolderContainer.tsx` does, gated on the
 `transferMove` feature flag, which renders Move in the Egress -> NetApp
-direction only).
+direction only. New screen (`transfer-materials-v1/`): `TransferControls.tsx`
+renders the `Move selected` button only when the source is Egress. Either way,
+Move is available in the Egress -> NetApp direction only.
 
 ## Project Structure
 
@@ -48,6 +59,12 @@ e2e/pw/
     register-case-state.ts            # Shared state file paths for the setup project
     types.ts                          # TypeScript type definitions
   pages/                              # Page Object Models
+    getTransferMaterialsTab.ts        # Picks the screen POM from TRANSFER_MATERIALS_V1
+    TransferMaterialsTabApi.ts        # Screen-agnostic Transfer Materials contract
+    TransferMaterialsTab.ts           # Old-screen implementation
+    TransferMaterialsTabV1.ts         # New-screen (v1) implementation
+    TransferDestinationPage.ts        # New-screen destination-tree page (pick folder + confirm)
+    ...                               # CaseSearch, SearchResults, CaseManagement, ActivityLog, login/connect POMs
   tests/
     register-case.setup.ts            # Setup project: register + connect once per run
     register-case.teardown.ts         # 24h workspace sweep, run after register-case-tests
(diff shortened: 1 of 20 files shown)


e2e/pw/pages/getTransferMaterialsTab.ts
```diff
@@ -0,0 +1,19 @@
...
+import { TransferMaterialsTabApi } from "./TransferMaterialsTabApi";
+
+export type { TransferMaterialsTabApi };
+
+/**
+ * Build the Transfer Materials page object for the active screen — the
+ * `TRANSFER_MATERIALS_V1` switch decides. Specs call this to stay
+ * screen-agnostic.
+ */
+export function getTransferMaterialsTab(page: Page): TransferMaterialsTabApi {
```
HasanCPS, line 14: ## The v1 switch only reaches 2 of the 11 specs

The factory works, but almost nothing calls it. Only `login-only.spec.ts:30` and `seed-netapp-fixture.setup.ts:85` build the tab this way. The other nine specs still do `const transferTab = new TransferMaterialsTab(page)` directly, so they pin themselves to the old screen no matter what the flag says:

```
egress-to-netapp-copy.spec.ts:29          egress-to-netapp-copy-default.spec.ts:31
egress-to-netapp-move.spec.ts:32          egress-to-netapp-move-default.spec.ts:31
egress-to-netapp-copy-large.spec.ts:32    egress-to-netapp-copy-large-default.spec.ts:34
egress-to-netapp-copy-multifile.spec.ts:32
netapp-to-egress-copy.spec.ts:31          netapp-to-egress-copy-default.spec.ts:28
```

What that does to a run with `TRANSFER_MATERIALS_V1=true` against a v1 environment: the seed setup goes through the factory, correctly drives v1, and passes. Then `netapp-to-egress-copy-default.spec.ts` runs the old page object against the v1 DOM. `TransferMaterialsTab.selectAction` looks for testid `netapp-inset-text` and `confirmTransfer` waits on testid `div-modal`. Neither exists on the new screen, which renders `Copy selected` / `Move selected` buttons and replaces the modal with the destination page. The spec dies on a 30 second `waiting for getByTestId('div-modal')` timeout, which tells the reader nothing about the actual cause.

Someone reading `.env.template:37` will reasonably expect one env var to switch the whole suite. Could you swap the remaining nine `new TransferMaterialsTab(page)` calls over to `getTransferMaterialsTab(page)`?

e2e/pw/README.md
```diff
@@ -189,6 +206,37 @@ standard auth set.
...
+redesigned screen. It is a **selector only** — it must match what the
+environment renders for `E2E_AD_USER`, or the specs drive the wrong screen and fail.
+
+Specs never branch on this flag directly. They build the Transfer Materials
+page object via `getTransferMaterialsTab(page)`, which reads the switch and
+returns either `TransferMaterialsTab` (old screen) or `TransferMaterialsTabV1`
+(new screen). Both implement the shared `TransferMaterialsTabApi` contract, so
+a spec depends only on that surface and stays screen-agnostic. The new screen
+adds a destination-tree step after Copy/Move — driven by `TransferDestinationPage`
+— because it has no confirm modal. A spec that must diverge (e.g.
```
HasanCPS: ## The README describes a spec branch that does not exist

Two claims in this paragraph do not match the code. "Specs never branch on this flag directly. They build the Transfer Materials page object via `getTransferMaterialsTab(page)`" is true for two specs and false for nine ([see comment 1](https://github.com/CPS-Innovation/Large-and-Complex-Cases/pull/496/changes#r3586474043)). And the named example does not do what it says: grepping `transferMaterialsV1` across `e2e/pw` hits only `helpers/env-config.ts` and `pages/getTransferMaterialsTab.ts`. No spec reads it, `netapp-to-egress-copy-default` included.

This matters more than a normal doc nit because the README is the thing that would otherwise tell someone the flag is only partly wired. Right now it reassures them it is fully wired. Once comment 1 is fixed the first claim becomes true; the `netapp-to-egress-copy-default` example still needs to go or be replaced with a real one. Can we bring this paragraph back in line with the code?

e2e/pw/pages/TransferMaterialsTabV1.ts
```diff
@@ -0,0 +1,177 @@
...
+    );
+
+    await Promise.race([
+      successBanner.waitFor({ state: "visible", timeout }),
+      this.page.waitForURL(TRANSFER_ERROR_ROUTE, { timeout }),
+    ]);
+
+    if (TRANSFER_ERROR_ROUTE.test(this.page.url())) {
+      const detail = await this.page
+        .locator("main")
```
HasanCPS, line 167: ## `waitForTransferComplete` can read the previous screen's `main`

`waitForURL` resolves the moment the URL changes, which is before React has rendered the error page. The `innerText()` on the next line then resolves against whichever `<main>` is mounted at that instant, and since every page wraps content in `PageContentWrapper` (which renders `<main id="main-content">`), it happily returns the *transfer materials* main instead of the error page's.

Concretely, on a re-run of the seed with the flag on: the duplicate is rejected, the app navigates to `/case/<id>/case-management/transfer-errors`, and `detail` comes back without "Some files already exist in the destination folder". The thrown message then fails the `/files? already exist/i` check at `seed-netapp-fixture.setup.ts:107`, so the seed hard-fails instead of being the idempotent no-op it is documented as.

The old implementation at `TransferMaterialsTab.ts:89-98` does not have this hole, because it waits for the error heading to be visible before reading `main`. Could you wait on the error heading (or any error-page element) before reading the text here?

e2e/pw/pages/TransferMaterialsTabV1.ts
```diff
@@ -0,0 +1,177 @@
...
+  /**
+   * Verify a file landed in the shared drive: switch to it as source and wait
+   * for the file's row checkbox to attach (attached-not-visible, as in
+   * selectNetAppFileByExactName). Timeout covers slow load + indexing lag.
+   */
+  async verifyNetAppContainsFile(
+    fileName: string,
+    timeout: number = 60_000,
+  ): Promise<void> {
+    await this.switchToNetAppSource();
```
HasanCPS, line 197: ## `verifyNetAppContainsFile` assumes the shared drive is never already the source

`switchToNetAppSource` clicks a button named "View Shared Drive", but that is one toggle whose label flips rather than two separate buttons. `TransferControls.tsx:38` renders "View Shared Drive" when the source is Egress and "View Egress" when it is already the shared drive. So when the shared drive is already the source, this clicks a button that is not on the page.

That happens in any NetApp to Egress spec, and on any second call to `verifyNetAppContainsFile` against the same page. You get a 30 second locator timeout reporting a missing button, instead of either the pass you wanted or the honest "file is not in the shared drive" failure. Can we make the toggle conditional on the current source, so it only clicks when a switch is actually needed?

e2e/pw/pages/TransferMaterialsTabV1.ts
```diff
@@ -0,0 +1,177 @@
...
+  async selectReverseAction(action: "Copy" | "Move"): Promise<void> {
+    await this.clickTransferControl(action);
+  }
+
+  /**
+   * Confirm the transfer. No modal on the new screen: Copy/Move already
+   * navigated to the destination tree — pick the first selectable folder (the
+   * connected root) and click the `<Copy|Move> to <folder>` button.
+   */
+  async confirmTransfer(): Promise<void> {
```
HasanCPS: ## `confirmTransfer()` depends on call order that the shared contract does not express

`lastAction` is written only by `clickTransferControl` (line 39) and defaults to "Copy" (line 27). `confirmTransfer()` reads it to build the destination-page regex, `^Copy to ` or `^Move to `, matching the button the SPA renders as `${transferAction} to ${selectedNode.name}`. So `confirmTransfer()` is only correct if the same instance just ran `selectAction`.

The old implementation has no such requirement. Its modal is action-agnostic, so `TransferMaterialsTabApi` advertises `confirmTransfer()` as something you can call on its own. Any caller who takes the interface at its word and reaches `confirmTransfer()` without that same instance having run `selectAction` first, say a Move spec that rebuilds the page object after a `page.reload()`, clicks a "Copy to ..." button that does not exist on a Move flow. The failure names the wrong action and gives no hint that state was lost.

No current caller does this, so it is latent rather than broken. But it is the kind of thing that bites whoever writes the next spec. Could you pass the action in as `confirmTransfer(action)` on the contract instead of carrying it in hidden instance state?

e2e/pw/pages/TransferDestinationPage.ts
```diff
@@ -0,0 +1,104 @@
...
+      exact: true,
+    });
+  }
+
+  /** Expand a folder node (no-op if already expanded). */
+  async expandFolder(folderName: string): Promise<void> {
+    const toggle = this.folderNode(folderName)
+      .locator("xpath=..")
+      .getByRole("button", { name: "plus" });
+    if ((await toggle.count()) > 0) {
```
HasanCPS: ## `expandFolder` silently does nothing while children are still loading

`count()` does not auto-wait. It snapshots the DOM and returns 0, so if the node is not there yet the `if` is skipped and the method returns as though it succeeded.

Walk through `chooseFolder("Copy", ["root", "4. Served Evidence", "target"])`. Expanding "root" clicks the plus, but the child nodes arrive via `TreeView`'s async `onLoadChildren`, which is a real `getNetAppFolders` / `getEgressFolders` network call. The next `expandFolder("4. Served Evidence")` runs before that fetch settles, matches 0 elements, and no-ops. The subtree never opens, and the test dies later on `selectFolder(target)` with a "locator resolved to 0 elements" timeout that points at the target folder rather than the ancestor that never expanded.

Worth noting this only bites once `chooseFolder` has a caller. Could you use `expect(toggle).toBeVisible()` or a `waitFor` here instead of the `count()` gate, so a missing node either waits or fails honestly?

e2e/pw/pages/TransferDestinationPage.ts
```diff
@@ -0,0 +1,104 @@
...
+      .click();
+  }
+
+  async cancel(): Promise<void> {
+    await this.page.getByRole("button", { name: "Cancel" }).click();
+  }
+
+  /** Expand ancestors, select the final folder, and confirm. `folderPath` runs
+   * from a tree root down to the target. */
+  async chooseFolder(
```
HasanCPS, line 97: ## Half of `TransferDestinationPage` has no caller

The only consumer is `TransferMaterialsTabV1.confirmTransfer` (line 121), and it uses just `waitForLoaded`, `selectFirstSelectableFolder` and `confirm`. That leaves `chooseFolder` unused, `expandFolder` and `selectFolder` unreachable except through `chooseFolder`, and `cancel` with no caller either. Roughly half of the 104 lines.

The problem is not the line count, it is that no test run will ever validate this code, including the fragile `locator("xpath=..")` parent hop and the lowercased-aria-label assumption. It rots quietly against DOM changes, and the next reader assumes these paths are proven because they are sitting in a page object next to code that is. Can we cut these until a spec actually needs a nested destination folder, and bring them back with the spec that needs them?

e2e/pw/pages/BaseTransferMaterialsTab.ts
```diff
@@ -0,0 +1,110 @@
...
+  async navigateToFolder(folderName: string): Promise<void> {
+    await this.page.getByRole("button", { name: folderName }).click();
+  }
+
+  /**
+   * Sort the NetApp panel by last-modified date descending (two header clicks).
+   * Best-effort: the sort doesn't reliably toggle to descending, so a flaky/slow
+   * header is tolerated and the panel is left unsorted.
+   */
+  async sortNetAppByDateDescending(): Promise<void> {
```
HasanCPS, line 49: ## `sortNetAppByDateDescending` is dead while the spec that needs it hand-rolls the same thing

Grepping the name across the repo returns three hits, all definitions or declarations, no invocations. Meanwhile `netapp-to-egress-copy.spec.ts:43-49` still does the identical two-click date sort inline against its own hardcoded `getByTestId("netapp-table-wrapper").getByRole("button", { name: "Last modified date" })`.

That hardcoded test id is the tell: the base method uses `this.netAppWrapperTestId`, which v1 overrides to `"shared drive-table-wrapper"`, so the spec's inline version is the one that breaks on the new screen. It is also declared on `TransferMaterialsTabApi` (line 15), which forces both implementations to satisfy an operation nothing calls. When the sort header's accessible name changes, someone will fix the base method, see no test change, and the spec will keep failing.

Could you have the spec call `transferTab.sortNetAppByDateDescending()` and drop the inline block?

e2e/pw/pages/TransferMaterialsTabV1.ts
```diff
@@ -0,0 +1,177 @@
...
+    try {
+      await checkbox.waitFor({ state: "attached", timeout: 120_000 });
+    } catch {
+      throw this.fixtureMissingError(fileName, "shared drive panel");
+    }
+    await checkbox.scrollIntoViewIfNeeded().catch(() => {});
+    await checkbox.check({ force: true });
+  }
+
+  async selectAction(action: "Copy" | "Move"): Promise<void> {
```
HasanCPS: ## `selectAction` and `selectReverseAction` are the same method twice

Both are a one-line delegation to `clickTransferControl(action)`, byte for byte identical. That is honest, because the split only exists for the old screen's two per-panel inset controls and the new screen has a single shared control.

Keeping both on `TransferMaterialsTabApi` (lines 18-19) means every future implementation carries a distinction that only one screen has, and a spec author reading the interface cannot tell which one to call for the new screen, because the answer is "either, they are the same". Could we collapse this to a single `selectAction(action, direction?)`, or let the source panel imply the direction?

HasanCPS (approved):
LGTM!

PR #499, opened 2026-07-15
move for manage materials becomes synchronous

bridgesr:
# PR checklist

## Correctness

## Keeping it simple

### If you touched the UI
- [ ] Ran the e2e/ suite locally (the PR build only runs the mocked tests; the full suite runs post-merge and gates the deploy)
- [ ] New config or feature flags are wired into the VITE build variables, not just added in code
- [ ] Data fetches handle the loading, empty and error states, not just success, and match how the rest of the app already does it
- [ ] Ran prettier (npm run prettier:format), since CI lints but doesn't check formatting
- [ ] Checked accessibility (keyboard nav, screen reader, sensible labels), since the Lighthouse run is manual and won't flag it for you


backend/CPS.ComplexCases.API/Functions/InitiateBatchMove.cs
```diff
@@ -41,21 +62,307 @@ public async Task<IActionResult> Run(
...
+        {
+            _logger.LogWarning("Validation failed for {OperationName}. CorrelationId: {CorrelationId}, Errors: {Errors}",
+                nameof(InitiateBatchMove), context.CorrelationId, batchRequest.ValidationErrors);
+            return new BadRequestObjectResult(batchRequest.ValidationErrors);
+        }
+
+        var request = batchRequest.Value;
+        _initializationHandler.Initialize(context.Username, context.CorrelationId, request.CaseId);
+
+        var pathError = await ValidatePathsAsync(request, context.CorrelationId);
```
HasanCPS, line 80: ## The active-transfer conflict check is gone

On `main`, the old function called `CheckActiveTransferConflictAsync` and returned a 409 ("A case-wide file transfer is in progress") before doing anything. That helper read `caseMetadata.ActiveTransferId` and then asked the durable orchestration entity whether the transfer was actually `InProgress`. The rewrite loads case metadata in `ValidatePathsAsync` but only uses it for the folder-path and prefix checks. Nothing looks at an active transfer anymore.

So if someone starts a batch move while a case-wide transfer for the same case is still running, the ONTAP relocate renames the source files out from under it, and the transfer copies partial data or fails. This is the one I'd want sorted before merge. The manage-materials lock a few lines down doesn't cover it, since that guards path conflicts between manage-materials operations, not a case-wide transfer.

Heads up that this isn't a straight re-add: the old check needed the `DurableTaskClient` to read the entity state, and this function doesn't have one. So we need another way to tell whether the transfer is genuinely in flight, whether that's trusting `ActiveTransferId.HasValue`, or querying transfer status through a service. Worth a quick chat on which.

backend/CPS.ComplexCases.API/Functions/InitiateBatchMove.cs
```diff
@@ -41,21 +62,307 @@ public async Task<IActionResult> Run(
...
+        var arg = _ontapArgFactory.CreateMaterialRenameArg(bearerToken, volumeUuid, operation.SourcePath, destinationPath);
+
+        try
+        {
+            var result = await _ontapHttpClient.RenameMaterialAsync(arg);
+            return CreateItemResult(operation, destinationPath, MapRenameResultToStatus(result), result.Success ? null : result.ErrorMessage);
+        }
+        catch (Exception ex) when (IsAuthException(ex))
+        {
+            throw;
```
HasanCPS, line 209: ## An auth failure part-way through the batch leaves the moved items unaudited

```csharp
catch (Exception ex) when (IsAuthException(ex))
{
    throw;
}
```
Say a batch of three: items 1 and 2 relocate fine, then item 3's `RenameMaterialAsync` throws `OntapUnauthorizedException` because the token expired mid-request. The rethrow propagates out of `ExecuteMovesAsync` (no catch there) and out of `Run`. The `finally` releases the lock, but `WriteMoveActivityLogAsync` on line 106 never runs and the middleware returns 401.

Items 1 and 2 are physically moved on NetApp, the caller sees an auth error, and the activity log has no record of the two that moved. Rethrowing an auth failure instead of recording it per item is a reasonable call, so I'm not asking to change that. The gap is the audit: the items that already moved should be logged before the exception bubbles. Worth noting `MaterialBatchRename` has the same rethrow-then-audit ordering (line 147), so whatever we decide probably wants to apply to both.

backend/CPS.ComplexCases.API/Functions/InitiateBatchMove.cs
```diff
@@ -41,21 +62,307 @@ public async Task<IActionResult> Run(
...
+        {
+            return null;
+        }
+
+        _logger.LogWarning("Source paths outside case folder: {Paths}. CorrelationId: {CorrelationId}",
+            invalidPaths, correlationId);
+        return new BadRequestObjectResult(new[] { $"The following source paths are not within the case's NetApp folder: {string.Join(", ", invalidPaths)}" });
+    }
+
+    private async Task<List<MoveNetAppBatchItemResult>> ExecuteMovesAsync(
```
HasanCPS: ## Nothing stops two operations in one batch landing on the same destination

The old code grouped operations by `DestinationPrefix + DestinationFileName` and returned a 400 up front when two collided (`main`, deleted `InitiateBatchMove.cs:142`). The current validator (`NetAppBatchValidatorBase` and `NetAppBatchCopyValidationRules.GetCrossFieldErrors`) only looks at source paths for duplicates and overlaps, and per-op source-vs-dest. It never compares destinations across operations.

So move `folderA/report.pdf` and `folderB/report.pdf` into the same prefix in one batch: sources differ, so validation passes, the first relocates, and the second hits an ONTAP 409 that comes back as `Conflict`. The batch reports `PartiallyCompleted` and one file quietly stays put, where the old behaviour was a clear up-front rejection.

Reasonable people could argue per-item `Conflict` is fine, but it's a behaviour change from a hard 400, so flagging it. If we want the old contract back, an inter-op destination-collision rule in the validator would do it and fail before we touch NetApp.

backend/CPS.ComplexCases.API/Functions/InitiateBatchMove.cs
```diff
@@ -41,21 +62,307 @@ public async Task<IActionResult> Run(
...
+        var name = lastSlash >= 0 ? trimmed[(lastSlash + 1)..] : trimmed;
+        return isFolder ? $"{destinationPrefix}{name}/" : $"{destinationPrefix}{name}";
+    }
+
+    private static string? ValidateOperation(string sourcePath, string destinationPath, bool isFolder)
+    {
+        if (string.Equals(destinationPath, sourcePath, StringComparison.OrdinalIgnoreCase))
+        {
+            return "The item is already in the destination location.";
+        }
```
HasanCPS: ## Re-moving a folder that's already in place reports the batch as "Failed"

```csharp
if (string.Equals(destinationPath, sourcePath, StringComparison.OrdinalIgnoreCase))
{
    return "The item is already in the destination location.";
}
```

`ValidateOperation` returns this as `OperationResultStatus.Failed` (line 179). For materials it isn't reachable, since `GetCrossFieldErrors` already rejects source == dest at validation (rules line 45). For folders it is reachable: the folder branch of `GetCrossFieldErrors` only checks whether the destination sits inside the source, not whether they're equal. So moving folder `case/x/myfolder/` with destination prefix `case/x/` passes validation, reaches `ValidateOperation`, and comes back `Failed`.

`BuildBatchResponse` counts that toward `failed`, so a lone "already there" folder yields status `"Failed"`. A no-op re-submit reporting as an outright failure will confuse users and anyone reading the logs. "Already in place" wants to be its own outcome, not a failure.

backend/CPS.ComplexCases.API/Functions/InitiateBatchMove.cs
```diff
@@ -41,21 +62,307 @@ public async Task<IActionResult> Run(
...
+        };
+    }
+
+    private static MoveNetAppBatchResponse BuildBatchResponse(int totalRequested, List<MoveNetAppBatchItemResult> results)
+    {
+        var succeeded = results.Count(r => r.Status == OperationResultStatus.Moved);
+        var notFound = results.Count(r => r.Status == OperationResultStatus.NotFound);
+        var failed = results.Count(r => r.Status is OperationResultStatus.Failed or OperationResultStatus.Conflict);
+
+        var batchStatus = (succeeded > 0, failed > 0) switch
```
HasanCPS: ## Batch status ignores NotFound, so an all-not-found batch still reports "Completed"

```csharp
var batchStatus = (succeeded > 0, failed > 0) switch
{
    (false, false) => "NoOp",
    (true, false) => "Completed",
    (false, true) => "Failed",
    _ => "PartiallyCompleted",
};
```

`notFound` is counted just above but never used here, so 1 Moved and 2 NotFound resolves to `Completed`. The response body does carry `NotFound = 2`, so a client reading the counts is fine, but one keying off `Status` alone is told everything moved.

Lower priority than it first looks: this switch is copied verbatim from `MaterialBatchRename` (line 217), so it's an existing convention rather than something this PR introduces. Still, if we're touching it, folding `notFound` into the decision (any `succeeded` alongside any `notFound`/`failed` being `PartiallyCompleted`) would make `Status` mean what it says. Worth deciding for both endpoints together, not just here.

backend/CPS.ComplexCases.API/Functions/InitiateBatchMove.cs
```diff
@@ -41,21 +62,307 @@ public async Task<IActionResult> Run(
...
+        catch (Exception ex)
+        {
+            _logger.LogError(ex, "Unexpected error moving {SourcePath} to {DestinationPath}.", operation.SourcePath, destinationPath);
+            return CreateItemResult(operation, destinationPath, OperationResultStatus.Failed, ex.Message);
+        }
+    }
+
+    private async Task WriteMoveActivityLogAsync(MoveNetAppBatchDto request, List<MoveNetAppBatchItemResult> results, string userName)
+    {
+        var hasAuditableResult = results.Any(r => r.Status is OperationResultStatus.Moved or OperationResultStatus.NotFound);
```
HasanCPS, line 220: ## NotFound items get written into the audit log's action type

```csharp
var hasAuditableResult = results.Any(r => r.Status is OperationResultStatus.Moved or OperationResultStatus.NotFound);
```

`NotFound` counts as auditable and feeds `auditedPaths`, which drives `hasFolder` / `hasMaterial` and the `actionType`. A batch with one material moved and one folder that came back `NotFound` produces `ActionType.FolderAndMaterialMoved`, so the top-level action claims a folder moved when it didn't exist.

Two things keep this mild: the `details` block logs every item with its real `outcome` (so the NotFound folder shows as NotFound in the detail), and this is again copied straight from `MaterialBatchRename` (line 172). So the audit isn't false, just imprecise at the action-type level, and it's consistent with the rename endpoint. If we tighten it, dropping `NotFound` from `auditedPaths` and the action-type decision (while keeping it in `details`) is the change. Same "do it in both places" caveat as [this](https://github.com/CPS-Innovation/Large-and-Complex-Cases/pull/499/changes#r3594624355).

backend/CPS.ComplexCases.API/Functions/InitiateBatchMove.cs
```diff
@@ -41,21 +62,329 @@ public async Task<IActionResult> Run(
...
+
+        return new MoveNetAppBatchResponse
+        {
+            Status = NetAppBatchOutcome.ResolveStatus(succeeded, failed, notFound + alreadyInPlace),
+            TotalRequested = totalRequested,
+            Succeeded = succeeded,
+            NotFound = notFound,
+            Failed = failed,
+            Results = results,
+        };
```
HasanCPS, line 323: ## AlreadyInPlace is a clean fix, but it's invisible in the summary counts

The new `AlreadyInPlace` status sorts item 4 out properly: a re-move of a folder that's already there comes back as its own outcome and folds into the "skipped" bucket, so an all-already-there batch is `NoOp` instead of `Failed`. Good.

The gap is the response shape. `AlreadyInPlace` isn't counted in `Succeeded`, `NotFound`, or `Failed`, and there's no `AlreadyInPlace` (or `Skipped`) field. So for a batch that has one, `Succeeded + NotFound + Failed` no longer equals `TotalRequested`, and the only way a client learns an item was already in place is by walking `Results[]` and matching the status string. `Status` itself resolves correctly, so this is observability, not correctness. Either add a count alongside the others, or note in the DTO doc that the three counts don't sum to the total once items are skipped.

HasanCPS (approved):
LGTM!

PR #505, opened 2026-07-17
fix: egress search with space

bridgesr:
# PR checklist

## Correctness

## Keeping it simple


HasanCPS (approved):
LGTM!

PR #509, opened 2026-07-21
refactoring to reduce cognitive complexity

bridgesr:
# PR checklist

## Correctness

## Keeping it simple


backend/CPS.ComplexCases.FileTransfer.API/Durable/Activity/TransferFile.cs
```diff
@@ -136,152 +105,284 @@ public async Task<TransferResult> Run([ActivityTrigger] TransferFilePayload payl
...
+                    startTime);
 
-                telemetryEvent.FileSizeInBytes = totalSize;
-                telemetryEvent.TransferEndTime =
-                    result.SuccessfulItem?.EndTime ?? result.SkippedItem?.EndTime ?? DateTime.UtcNow;
-                telemetryEvent.IsSuccessful = result.IsSuccess || result.IsSkipped;
-                telemetryEvent.IsMultipart = result.IsSuccess && result.SuccessfulItem!.TotalPartsCount > 1;
-                telemetryEvent.TotalPartsCount = result.IsSuccess ? result.SuccessfulItem!.TotalPartsCount : 0;
+                ApplyTransferResultTelemetry(telemetryEvent, totalSize, result);
+                TrackTransferTelemetry(telemetryEvent);
```
HasanCPS: **Double telemetry on success:** `TrackTransferTelemetry` is called inside the try block (after `ApplyTransferResultTelemetry`) and then again unconditionally in `finally`. On the success path both fire, so App Insights gets a duplicate event per transfer. Remove the `TrackTransferTelemetry(telemetryEvent)` call inside the try block. Let `finally` be the single emit point for all paths.

backend/CPS.ComplexCases.FileTransfer.API/Durable/Orchestration/TransferOrchestrator.cs
```diff
@@ -52,326 +52,391 @@ public async Task RunOrchestrator(
...
+            await RetryTransientFailuresAsync(
+                context, input, transferEntity, cleanFiles, entityId, allResults, transferOrchestrationEvent, logger);
+
+            await DeleteSourceFilesIfMoveAsync(context, input);
+
+            await FinalizeAndLogCompletionAsync(context, input);
+
+            transferOrchestrationEvent.IsSuccessful = transferOrchestrationEvent.TotalFilesFailed == 0;
+            transferOrchestrationEvent.OrchestrationEndTime = context.CurrentUtcDateTime;
+            _telemetryClient.TrackEvent(transferOrchestrationEvent);
```
HasanCPS: **[Same ](https://github.com/CPS-Innovation/Large-and-Complex-Cases/pull/509/changes#r3674969467)double-telemetry pattern:** `_telemetryClient.TrackEvent(transferOrchestrationEvent)` fires at the end of the try block and again in `finally`. On success both execute. Remove the `TrackEvent` call inside the try block and let `finally` handle it. Set `IsSuccessful` before the `finally` runs (it already does), or move the success flag assignment into `finally` guarded by a check.

backend/CPS.ComplexCases.FileTransfer.API/Durable/Activity/TransferFile.cs
```diff
@@ -136,152 +105,284 @@ public async Task<TransferResult> Run([ActivityTrigger] TransferFilePayload payl
...
         }
-        catch (FileExistsException ex)
+        finally
         {
-            LogFileConflictTelemetry(payload);
-            telemetryEvent.ErrorCode = TransferErrorCode.FileExists.ToString();
-            telemetryEvent.ErrorMessage = ex.Message;
-            return CreateFailureResult(payload.SourcePath.FullFilePath ?? payload.SourcePath.Path,
-                TransferErrorCode.FileExists, MapUserMessage(TransferErrorCode.FileExists), ex);
+            telemetryEvent.TransferEndTime = DateTime.UtcNow;
```
HasanCPS: `finally` block  `telemetryEvent.TransferEndTime = DateTime.UtcNow` unconditionally overwrites the value that `ApplyTransferResultTelemetry` set from `result.SuccessfulItem?.EndTime`. The accurate end time from the actual transfer result is lost. Guard it: `telemetryEvent.TransferEndTime = telemetryEvent.TransferEndTime == default ? DateTime.UtcNow : telemetryEvent.TransferEndTime;`

backend/CPS.ComplexCases.FileTransfer.API/Durable/Activity/TransferFile.cs
```diff
@@ -136,152 +105,284 @@ public async Task<TransferResult> Run([ActivityTrigger] TransferFilePayload payl
...
+            {
+                Rethrow = false,
+                ErrorCode = TransferErrorCode.FileExists,
+                DiagnosticMessage = fileExists.Message,
+                LogFileConflict = true,
+                Exception = fileExists
+            };
+        }
+
+        if (ex is OperationCanceledException oce && !isCancellationRequested)
```
HasanCPS, line 290: The two `OperationCanceledException` branches (timeout vs real cancel) are separated by ~80 lines of other exception checks. Easy to miss the second one on a future read.  Add a brief comment above the second check: `// True cancellation (token was triggered) - rethrow to let the framework handle it.`

backend/CPS.ComplexCases.API/Functions/MaterialBatchRename.cs
```diff
@@ -91,93 +91,140 @@ public async Task<IActionResult> Run([HttpTrigger(AuthorizationLevel.Anonymous,
...
+        Guid volumeUuid)
+    {
+        var arg = _ontapArgFactory.CreateMaterialRenameArg(bearerToken, volumeUuid, operation.CurrentPath, operation.NewPath);
+
+        try
+        {
+            var result = await _ontapHttpClient.RenameMaterialAsync(arg);
+            return MapRenameResultToItemResult(operation, result, _logger);
+        }
+        catch (Exception ex) when (IsAuthException(ex))
```
HasanCPS, line 134: `RenameSingleOperationAsync` inner `catch (Exception ex) when (IsAuthException(ex)) { throw; }`. This empty catch-and-rethrow is needed to stop the generic `catch` from swallowing auth failures, but it looks accidental without context. Add a comment: `// Re-throw auth failures so the outer handler can audit already-renamed items before propagating.` |

backend/CPS.ComplexCases.API/CPS.ComplexCases.API.csproj
```diff
@@ -29,7 +29,7 @@
     <PackageReference Include="Microsoft.Azure.Functions.Worker.ApplicationInsights" Version="2.0.0" />
     <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.DurableTask" Version="1.1.7" />
     <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore" Version="2.1.0" />
-    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="2.0.0" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="2.0.2" />
```
HasanCPS: `Microsoft.Azure.Functions.Worker.Sdk` bump 2.0.0 to 2.0.2 is unrelated to the refactoring. Bundling dependency updates with behaviour-preserving refactors makes rollback harder if one causes issues. Move the package bump to its own commit or PR.

HasanCPS (approved):
LGTM!

PR #512, opened 2026-07-23
0 byte transfers no longer show as fail cases

bridgesr:
# PR checklist

## Correctness

## Keeping it simple

### If you touched the UI

### DIFF
--- a/backend/CPS.ComplexCases.FileTransfer.API/Durable/Helpers/TransferResultProcessor.cs
+++ b/backend/CPS.ComplexCases.FileTransfer.API/Durable/Helpers/TransferResultProcessor.cs
@@ -1,0 +1,51 @@
+using CPS.ComplexCases.FileTransfer.API.Durable.Payloads.Domain;
+using CPS.ComplexCases.FileTransfer.API.Durable.State;
+using CPS.ComplexCases.FileTransfer.API.Telemetry;
+using Microsoft.DurableTask;
+using Microsoft.DurableTask.Entities;
+
+namespace CPS.ComplexCases.FileTransfer.API.Durable.Helpers;
+
+
+public static class TransferResultProcessor
+{
+    public static async Task ProcessAsync(
+        TaskOrchestrationContext context,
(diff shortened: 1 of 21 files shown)


backend/CPS.ComplexCases.FileTransfer.API/Durable/Helpers/TransferResultProcessor.cs
```diff
@@ -0,0 +1,51 @@
...
+            {
+                await context.Entities.CallEntityAsync(
+                    entityId,
+                    isRetry ? nameof(TransferEntityState.AddFailedRetryItem)
+                            : nameof(TransferEntityState.AddFailedItem),
+                    result.FailedItem);
+
+                telemetryEvent.TotalFilesFailed++;
+            }
+        }
```
brad-cps, line 86: The three branches don't cover the whole space and anything that misses all of them is dropped with no record at all. I'd consider adding an `else` that at minimum records the unclassifiable result.

backend/CPS.ComplexCases.FileTransfer.API/Durable/Helpers/TransferResultProcessor.cs
```diff
@@ -0,0 +1,51 @@
...
+    public static async Task ProcessAsync(
+        TaskOrchestrationContext context,
+        EntityInstanceId entityId,
+        TransferResult[] results,
+        TransferOrchestrationEvent telemetryEvent,
+        bool isRetry = false)
+    {
+        foreach (var result in results)
+        {
+            if (result != null && result.IsSkipped && result.SkippedItem != null)
```
brad-cps: Minor point. The repeated `result != null` in every branch reads better as a `if (result is null) continue;` guard at the top of the loop... with a sequential `await` per result.

brad-cps (dismissed):
Approved with minor comment

brad-cps (approved):
LGTM. Some Cognitive Complexity warning, but guess these can be resolved separately.

PR #515, opened 2026-07-27
Task/fct2 20588 add validation enhancements to e2 e playwright tests

kmshrajcps:
# PR checklist

  ## Correctness

  ## Keeping it simple

  ### If you touched the UI
  - [ ] Ran the e2e/ suite locally (the PR build only runs the mocked tests; the full suite runs post-merge and gates 
  the deploy)


lilachdavis (approved):
LGTM

PR #520, opened 2026-07-27
Fix/exclude large move test from cicd pipeline

lilachdavis:
We don't want to run the large file Move tests on every run of the deployment pipeline.
This test is intended to run nightly on its own.

# PR checklist

## Keeping it simple


kmshrajcps (approved):
LGTM

PR #523, opened 2026-07-30
egress and netapp path fix

bridgesr:
# PR checklist

## Correctness

## Keeping it simple

### If you touched the UI

### DIFF
--- a/ui-spa/src/components/case-management/index.tsx
+++ b/ui-spa/src/components/case-management/index.tsx
@@ -100,6 +100,7 @@ const CaseManagementPage = () => {
               caseId={caseId}
               operationName={operationNameOrDefendantName}
               egressWorkspaceId={caseMetaData.egressWorkspaceId}
+              egressWorkspaceName={caseMetaData.egressWorkspaceName ?? ""}
               netAppPath={caseMetaData.netappFolderPath}
               activeTransferId={
                 routeState?.transferId ?? caseMetaData.activeTransferId ?? ""
--- a/ui-spa/src/components/case-management/index.tsx
+++ b/ui-spa/src/components/case-management/index.tsx
@@ -100,7 +100,10 @@ const CaseManagementPage = () => {
               caseId={caseId}
               operationName={operationNameOrDefendantName}
               egressWorkspaceId={caseMetaData.egressWorkspaceId}
-              egressWorkspaceName={caseMetaData.egressWorkspaceName ?? ""}
+              egressWorkspaceName={
+                caseMetaData.egressWorkspaceName ||
+                operationNameOrDefendantName
+              }
               netAppPath={caseMetaData.netappFolderPath}
               activeTransferId={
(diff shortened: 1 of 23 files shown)


ui-spa/src/components/case-management/transfer-materials-v1/TransferDestinationPage.tsx
```diff
@@ -118,7 +118,7 @@ const TransferDestinationPage: React.FC = () => {
     const folders = [
       {
         id: netAppPath,
-        name: `Shared Drive: ${operationName}`,
+        name: `Shared Drive: ${netAppPath}`,
```
renjithabby: is netAppPath always a single folder name or do we need to takeout the current foldername from the netAppPath? also do we need to remove the /

bridgesr: @renjithabby @HasanCPS I think this is a question for the business, I thought it's useful to actually show the full path that the case is connected to as it makes it explicit to the user. For example here we are connected to Delta/Evidence, maybe stripping the last "/" makes sense but I don't think removing the Delta/ does as that suggests its connected at the top level

Capture

ui-spa/src/components/case-management/index.tsx
```diff
@@ -100,6 +100,7 @@ const CaseManagementPage = () => {
               caseId={caseId}
               operationName={operationNameOrDefendantName}
               egressWorkspaceId={caseMetaData.egressWorkspaceId}
+              egressWorkspaceName={caseMetaData.egressWorkspaceName ?? ""}
```
HasanCPS: The schema marks this as `z.string().nullable()` and the backend has it as `string?`, so `null` is a valid value. If it comes through as null the UI will render `"Egress: "` with nothing after the colon. The old code used to fall back to `leadDefendantName` when operationName was null, so there was always something visible. We need to decide what happens here: a fallback value, a different label, or confirmation that this field is guaranteed non-null when a workspace ID exists. As it stands a user could see a broken-looking label in production.

ui-spa/src/mocks/data/caseMetaData.playwright.ts
```diff
@@ -2,6 +2,7 @@ import { type CaseMetaDataResponse } from "../../schemas";
 export const caseMetaDataPlaywright: CaseMetaDataResponse = {
   caseId: 12,
   egressWorkspaceId: "egress_1",
+  egressWorkspaceName: "Thunderstruck",
```
HasanCPS: Every mock sets `egressWorkspaceName: "Thunderstruck"` and `operationName: "Thunderstruck"` to the same value. If someone accidentally reads `operationName` instead of `egressWorkspaceName`, every test still passes. These tests give false confidence right now. Setting them to different values (e.g. `egressWorkspaceName: "Workspace-Alpha"`, `operationName: "Thunderstruck"`) would actually catch a wrong-field bug.

HasanCPS (approved):
LGTM!

PR #524, opened 2026-07-30
story/FCT2-21031-Maintenenace-Page

renjithabby:
# PR checklist

## Correctness

## Keeping it simple

### If you touched the UI


ui-spa/src/components/AppRoutes.tsx
```diff
@@ -36,7 +38,10 @@ const AppRoutes = () => {
         },
       });
     }
-  }, [featureFlagsData, featureFlags, dispatch]);
+    if (featureFlagsData?.maintenanceMode) {
```
HasanCPS: ####  Redirect fires on every render and there is no way to exit maintenance mode

The `useEffect` at line 41 navigates to `/maintenance` every time `featureFlagsData` changes and `maintenanceMode` is truthy. If the user is already on `/maintenance` and anything causes a re-render it will navigate again. More importantly, when the flag flips back to `false` the user stays stuck on `/maintenance` because there is nothing to send them back out.

```tsx
// Lines 41-43 — currently:
if (featureFlagsData?.maintenanceMode) {
  navigate("/maintenance", { replace: true });
}

// Suggested fix:
if (featureFlagsData?.maintenanceMode && location.pathname !== "/maintenance") {
  navigate("/maintenance", { replace: true });
}
if (!featureFlagsData?.maintenanceMode && location.pathname === "/maintenance") {
  navigate("/", { replace: true });
}
```

ui-spa/src/components/AppRoutes.tsx
```diff
@@ -46,6 +51,7 @@ const AppRoutes = () => {
         path="/case/:caseId/case-management"
         element={<CaseManagementPage />}
       />
+      <Route path="/maintenance" element={<MaintenancePage />} />
```
HasanCPS, line 61: #### `/maintenance` route is accessible to anyone regardless of whether maintenance mode is on

```tsx
<Route path="/maintenance" element={<MaintenancePage />} />
```

This sits outside `<ProtectedRoutes />` (line 56). Any user can navigate to `/maintenance` directly and see "Sorry this service is unavailable" even when the service is running fine. If someone bookmarks it or shares the link, they get a false outage message.

Either move it inside `<ProtectedRoutes />` with a conditional render based on the flag, or add a redirect inside `MaintenancePage` that sends users to `/` when `FEATURE_FLAG_MAINTENANCE_MODE` is `false`.

renjithabby: wouldnt  the fix above  (!featureFlagsData?.maintenanceMode && location.pathname === "/maintenance") {
  navigate("/", { replace: true });
}  fix this issue

renjithabby: the fist comment fix , will fix the second issue , added integration test for this scenario

HasanCPS (approved):
LGTM!

PR #525, opened 2026-07-30
include large move test in playwright config

lilachdavis:
Add a project to playwright config for the large move test


kmshrajcps (approved):
LGTM

PR #527, opened 2026-07-30
Story/FCT2-21088- Ui style update

renjithabby:
image

# PR checklist

## Correctness

## Keeping it simple

### If you touched the UI

### DIFF
--- a/ui-spa/src/components/govuk/LinkButton.module.scss
+++ b/ui-spa/src/components/govuk/LinkButton.module.scss
@@ -5,6 +5,8 @@
   border-width: 0;
   cursor: pointer;
   text-decoration: underline;
+  text-decoration-thickness: max(1px, 0.0625rem);
+  text-underline-offset: 0.1578rem;
   font-size: 1rem;
   width: fit-content;
   color: #1d70b8;
--- a/ui-spa/src/components/common/FolderPath.module.scss
+++ b/ui-spa/src/components/common/FolderPath.module.scss
@@ -30,6 +30,7 @@
     overflow-wrap: break-word;
     word-break: break-all;
     text-align: left;
+    color: #0b0c0c;
   }
 }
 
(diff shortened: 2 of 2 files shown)


ui-spa/src/components/govuk/LinkButton.module.scss
```diff
@@ -5,6 +5,8 @@
   border-width: 0;
   cursor: pointer;
   text-decoration: underline;
+  text-decoration-thickness: max(1px, 0.0625rem);
+  text-underline-offset: 0.1578rem;
```
HasanCPS, line 9: ### LinkButton underline changes are global — scope creep beyond the ticket

The `text-decoration-thickness` and `text-underline-offset` additions apply to **every** `LinkButton` across the app (14+ components use it), not just the folder path buttons. The ticket only asks for a color change on folder path link buttons. These underline tweaks might be the right GOV.UK alignment, but they should either:
- be scoped to `FolderPath.module.scss` if they're only meant for the folder breadcrumb, or
- be covered by a separate ticket / explicitly called out in the PR description if the intent is to align all LinkButtons to the GOV.UK link style.

Without that context, as a reviewer I can't tell if this is intentional or accidental.

```diff
 # Current change (global — affects all LinkButtons):
   text-decoration: underline;
+  text-decoration-thickness: max(1px, 0.0625rem);
+  text-underline-offset: 0.1578rem;
   font-size: 1rem;
```

```diff
 # If scoping to FolderPath only, move to FolderPath.module.scss instead:
   & button {
     overflow-wrap: break-word;
     word-break: break-all;
     text-align: left;
     color: #0b0c0c;
+    text-decoration-thickness: max(1px, 0.0625rem);
+    text-underline-offset: 0.1578rem;
   }
```

HasanCPS (approved):
LGTM!

PR #528, opened 2026-07-30
don't allow duplicate connections for a case

bridgesr:
# PR checklist

## Correctness

## Keeping it simple

### If you touched the UI


backend/CPS.ComplexCases.API/Functions/CreateNetAppConnection.cs
```diff
@@ -71,6 +72,15 @@ public async Task<IActionResult> Run([HttpTrigger(AuthorizationLevel.Anonymous,
             return new UnauthorizedResult();
         }
 
+        var folderPath = netAppConnectionRequest.Value.NetAppFolderPath;
+        var existingConnections = await _caseMetadataService.GetCaseMetadataForNetAppFolderPathsAsync([folderPath]);
+        var existingConnection = existingConnections.FirstOrDefault();
```
HasanCPS, line 79: #### Duplicate check misses legacy `bucket:path` rows

`CreateNetAppConnection.cs` only queries the plain folder path. If a legacy row is stored as `bucket:path`, the check won't find it and a second case will be allowed to connect to the same physical folder — exactly the bug the ticket is fixing.

The enrichment service already handles both formats. The creation endpoint should too.

```csharp
// Current — only checks the plain path
var folderPath = netAppConnectionRequest.Value.NetAppFolderPath;
var existingConnections = await _caseMetadataService.GetCaseMetadataForNetAppFolderPathsAsync([folderPath]);
var existingConnection = existingConnections.FirstOrDefault();
```

**Suggested fix:**

```csharp
var folderPath = netAppConnectionRequest.Value.NetAppFolderPath;
var bucketName = securityGroups.First().BucketName;
var lookupPaths = new[] { folderPath, $"{bucketName}:{folderPath}" };
var existingConnections = await _caseMetadataService.GetCaseMetadataForNetAppFolderPathsAsync(lookupPaths);
var existingConnection = existingConnections.FirstOrDefault();
```

The `securityGroups` variable is already in scope a few lines above, so no extra call is needed.

The conflict test should also be updated to verify both lookup variants.

backend/CPS.ComplexCases.API/Functions/CreateNetAppConnection.cs
```diff
@@ -71,6 +72,15 @@ public async Task<IActionResult> Run([HttpTrigger(AuthorizationLevel.Anonymous,
...
+
+        if (existingConnection != null)
+        {
+            return new ConflictObjectResult($"Folder path '{folderPath}' is already connected to case {existingConnection.CaseId}.");
+        }
```
HasanCPS, line 87: #### No logging when a duplicate connection is rejected

A conflict rejection is a meaningful event (someone tried to connect to an already-taken folder). Adding a log line helps with debugging and audit trails, especially for support tickets.

```csharp
// Current
if (existingConnection != null)
{
    return new ConflictObjectResult($"Folder path '{folderPath}' is already connected to case {existingConnection.CaseId}.");
}
```

**Suggested fix:**

```csharp
if (existingConnection != null)
{
    _logger.LogWarning(
        "Duplicate NetApp connection attempt: folder path '{FolderPath}' is already connected to case {ExistingCaseId}. Requested by case {RequestedCaseId}.",
        folderPath, existingConnection.CaseId, netAppConnectionRequest.Value.CaseId);
    return new ConflictObjectResult($"Folder path '{folderPath}' is already connected to case {existingConnection.CaseId}.");
}
```

backend/CPS.ComplexCases.API/Functions/CreateNetAppConnection.cs
```diff
@@ -71,6 +72,15 @@ public async Task<IActionResult> Run([HttpTrigger(AuthorizationLevel.Anonymous,
             return new UnauthorizedResult();
         }
 
+        var folderPath = netAppConnectionRequest.Value.NetAppFolderPath;
+        var existingConnections = await _caseMetadataService.GetCaseMetadataForNetAppFolderPathsAsync([folderPath]);
+        var existingConnection = existingConnections.FirstOrDefault();
+
+        if (existingConnection != null)
+        {
+            return new ConflictObjectResult($"Folder path '{folderPath}' is already connected to case {existingConnection.CaseId}.");
```
HasanCPS: #### Conflict response body leaks the existing case ID to the caller

The `409 Conflict` response body includes the case ID of the other case: `"Folder path '...' is already connected to case 12345."` Depending on the security model, the requesting user may not have permission to know which case holds that folder. Consider whether the error message should omit the case ID.

```csharp
return new ConflictObjectResult($"Folder path '{folderPath}' is already connected to case {existingConnection.CaseId}.");
```

**Potential alternative:**

```csharp
return new ConflictObjectResult($"Folder path '{folderPath}' is already connected to another case.");
```

HasanCPS (approved):
LGTM!

PR #530, opened 2026-07-31
Task/fct2 21022 upgrade UI outdated node version

lilachdavis:
# Changes made
* Update UI deployment pipeline to set the webapp runtime to node 24-lts
* Update pipeline tasks to use node 24

# Background
The ITHC for Register a Case flagged our usage of the deprecated node 20 as a medium-severity issue.
The same applies to the Transfer materials solution.

# Testing
I manually updated the stack settings of the dev instance to Node 24, and confirmed that the webapp continued operating normally.
image

# PR checklist
## Correctness

## Keeping it simple


HasanCPS (approved):
LGTM!

PR #534, opened 2026-08-07
add telemetry filter for the /status endpoint

lilachdavis:
Add and register an AppInsights telemetry processor that filters out worker-generated telemetry for the Status health check endpoint.
There is currently no way to filter out host-generated telemetry by endpoint. This will be handled at the platform level using a Log Analytics workspace transformation Data Collection Rule (DCR).

**Without filter**, e2e transaction logs for a single Status request:
image
**With filter applied**:
image

# PR checklist
## Correctness

## Keeping it simple

- [ ] Ran dotnet format, since CI builds and tests but doesn't check formatting - **Running dotnet format from the `backend` folder changed 223 files. I didn't push these changes here. This would need to be a separate PR, just to consolidate formatting**


HasanCPS (approved):
LGTM!

PR #535, opened 2026-08-07
Fix Egress token expiry and surface transfer failures in e2e tests

kmshrajcps:
Egress token expiry silently disabled a soak assertion. Runs last ~an hour and outlive the  token. listEgressWorkspaceFilesByFolderId used a token minted at run start and returned
  [] on any non-OK response — so a 401 read as "file not in the folder", and the "file      
  removed from Egress after a move" check passed without checking anything.

  Changes
  
  - listEgressWorkspaceFilesByFolderId routes through fetchWithTokenRefresh when given a
  service account: a 401 re-authenticates and retries. Accepts a mutable token ref as well  
  as a string, so the refreshed token is written back and callers refresh once rather than  
  per call.
  - It now throws on non-OK instead of returning [], so auth failures can't look like a     
  deleted file. Teardown catches for itself to stay best-effort.
  - transfer-verify holds its token in a ref and passes the service account through,        
  replacing a cache that could never refresh.
  - Soak spec asserts failedFiles is zero and names the failed items. The poll loop exits on  PartiallyCompleted too, so dropped files previously surfaced as "file still exists in     
  Egress" — the symptom, not the cause.
  - Adds rebuild-default-fixture and seed-fixture, two opt-in projects for restoring a      
  missing DEFAULT_WORKSPACE_ID and seeding a workspace + case from scratch.

  Verification: move-large-test against staging, 1 passed — exercises both changed call     
  sites. The 401-refresh path itself is unverified; a 1.2-minute run can't expire a token,  
  so only the full soak will exercise it.


e2e/pw/move-soak/tests/egress-to-netapp-move-soak.spec.ts
```diff
@@ -127,6 +127,19 @@ test.describe("Move Soak Tests", () => {
         contentType: "application/json",
       });
 
+      // The poll loop exits on PartiallyCompleted and Failed as well as
+      // Completed, so a transfer that dropped files still reaches the
+      // verification steps below. A failed move correctly leaves its source
+      // file in Egress, so the first thing to break is the "removed from
+      // Egress" assertion - which reports the symptom and hides the cause.
+      // Assert here, after the attachments so the status is still captured.
+      expect(
```
lilachdavis: Any 'expect' blocks should come later than step 4. Cleanup transfer
`await harness.clearTransfer(transfer.id);` (line 165)
Otherwise, the test exists while the transfer is still in an active state, which blocks other users from performing any further actions on the case.

lilachdavis (changes requested):
Please see comment

e2e/pw/move-soak/tests/egress-to-netapp-move-soak.spec.ts
```diff
@@ -155,7 +155,16 @@ test.describe("Move Soak Tests", () => {
       expect(status).toBeTruthy();
 
       expect(status?.status).toBe("Completed");
-      expect(status?.failedFiles).toBe(0);
+      // Name the files the backend rejected: without them a partial transfer
+      // reports only "expected 0, received 2", and the next failure is
+      // "file still exists in Egress" - which is correct for a failed move
+      // and points away from the cause.
+      expect(
```
lilachdavis, line 160: If `status != "Completed"` execution stops at the previous assertion and this 3rd `expect()` never runs.
Only a `"PartiallyCompleted"` transfer contains any failedFiles.
This should be placed before `expect(status?.status).toBe("Completed");`

The issue with "file still exists in Egress" occurs after a Move has completed successfully.
When a move fails, we never get to the stage of verifying that files had been deleted from source.

lilachdavis (changes requested):

e2e/pw/move-soak/tests/egress-to-netapp-move-soak.spec.ts
kmshrajcps: Good catch,  failedFiles now asserts before the status check, so a PartiallyCompleted run reports the  
  file names instead of stopping at expected "Completed", received "PartiallyCompleted".

lilachdavis (approved):
LGTM

PR #542, opened 2026-08-14
disable egress source folder cleanup for debugging

lilachdavis:
To investigate if files are indeed failing to be deleted from Egress after move and eliminate the possibility that it's a timing issue with the test itself - I disabled the clean up jobs that removes any files in the egress source folder.


HasanCPS (approved):
LGTM!

PR #545, opened 2026-08-18
Feature/fct2 21384 selectable bucket

bridgesr:
# PR checklist

## Correctness

## Keeping it simple

## Code Quality
- [x] Pre-commit hooks were run for all commits in this PR

### If you touched the UI


backend/CPS.ComplexCases.API/Functions/SearchNetAppFolders.cs
HasanCPS: The `MaterialBatchRename.cs` file also needs to be updated to follow this pattern. The corresponding test files (`MaterialRenameTests.cs` and `MaterialBatchRenameHelperTests.cs`) should also be updated.

backend/CPS.ComplexCases.API/Functions/ListNetAppFiles.cs
```diff
@@ -45,9 +48,17 @@ public async Task<IActionResult> Run([HttpTrigger(AuthorizationLevel.Anonymous,
         var take = int.TryParse(req.Query[InputParameters.Take], out var takeValue) ? takeValue : 100;
         var path = req.Query[InputParameters.Path];
 
-        var securityGroups = await _securityGroupMetadataService.GetUserSecurityGroupsAsync(context.BearerToken);
```
HasanCPS: The ticket states that "list and browse endpoints need a bucket before the case has a persisted value (i.e. during the connect wizard)".

`ListNetAppFolders` correctly supports both the `caseId` and `bucketName` query parameters, whereas `ListNetAppFiles` only supports `caseId`.

As a result, if a user needs to preview files during the connect wizard before a case has a persisted bucket, `ResolveBucketAsync` is called with `(bearerToken, null, null)` and silently falls back to `.First()`.

This reintroduces the ordering bug for this code path, as bucket selection becomes dependent on the order of the returned results rather than an explicitly specified bucket.

Also add the corresponding `[OpenApiParameter]` attribute and update `ListNetAppFilesTests`.

HasanCPS (approved):
LGTM!

